PR #7, opened 2025-05-16
build: deploy examples to GitHub pages on main runs

spike-rabbit:
Publish the documentation on main runs.

See for a working version: https://spike-rabbit.github.io/element/components/
Examples are broken because of absolute link, but they should be working with element.siemens.io.

Code is on the main branch.

---

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -98,7 +98,25 @@ jobs:
         with:
           name: pages
           path: dist/design
-# TODO: build design system
+
+  publish-documentation:
+    if: github.ref_name == github.event.repository.default_branch
+    runs-on: ubuntu-22.04
+    permissions:
+      pages: write
+      id-token: write
+    needs:
+      - documentation
+    steps:
+      - uses: actions/configure-pages@v5
+      - uses: actions/download-artifact@v4
+        with:
+          name: pages
+          path: dist/design
+      - uses: actions/upload-pages-artifact@v3
+        with:
+          path: 'dist/design'
+      - uses: actions/deploy-pages@v4
 # TODO: add release job
 # TODO: build PR previews
 # TODO: e2e test


.github/workflows/build.yaml
```diff
@@ -98,7 +98,25 @@ jobs:
         with:
           name: pages
           path: dist/design
-# TODO: build design system
+
+  publish-documentation:
+    if: github.ref == 'refs/heads/main'
```
fh1ch, line 103: It probably makes sense if we derive the default branch directly from GitHub, so it's in-sync with all of our protection and deployment rules (see also https://stackoverflow.com/a/69995861):

```suggestion
    if: github.ref_name == github.event.repository.default_branch
```

.github/workflows/build.yaml
```diff
@@ -98,7 +98,25 @@ jobs:
...
+    steps:
+      - uses: actions/configure-pages@v5
+      - uses: actions/download-artifact@v4
+        with:
+          name: pages
+          path: dist/design
+      - uses: actions/upload-pages-artifact@v3
+        with:
+          path: 'dist/design'
+      - uses: actions/deploy-pages@v4
```
fh1ch, line 119: Probably more of a follow-up topic, but we might want to define an envrionment here at some point: https://github.com/actions/deploy-pages?tab=readme-ov-file#usage.

fh1ch (changes requested):
@spike-rabbit thanks a ton for the merge request 🙇 

Just one nit from my end, otherwise perfectly fine 🏓

.github/workflows/build.yaml
spike-rabbit: yes, once we can :laughing:

fh1ch (approved):
@spike-rabbit thanks a ton for the rework, all good now 💯 

LGTM 👍

PR #16, opened 2025-05-20
feat(element-translation-ng): move key extraction CLI into a separate package

spike-rabbit:
The key extraction must be separated for internal dependency clearing. As the key extraction is not included in an app bundle, its dependencies must not be cleared.
Moving it into its own package allows apps to install it as a dev dependency and therefore exclude it from clearing.

---


timowolf (approved):
Many thanks. This change is a very good idea.

PR #17, opened 2025-05-20
fix(live-preview): accept legacy component loader import

spike-rabbit:
To avoid breaking changes when re-exporting, we need to accept the old `@simpl/live-preview` package name.
Otherwise, we cannot re-export without a breaking change.

---


timowolf (approved):
LGTM

PR #31, opened 2025-05-26
build(github): use latest lts image

spike-rabbit:
Without this, the action will fail.
Don't really know why.

---


timowolf (changes requested):
Let us wait until the pilings are up and running again

timowolf:
@spike-rabbit pipeline seems to be up and running again. I would recommend do close this. What do you think?

spike-rabbit:
Let see if we may need this in the future.

PR #38, opened 2025-05-27
build(element.io): use uv for docs-builder plugins

spike-rabbit:
Removes poetry leftovers missed on initial migration.
With this change, we consistently use the same backends and
tools for the entire python parts.

---


fh1ch (approved):
@spike-rabbit huge thanks for the nice work here and thanks @max-wittig for the Python expertise 💚 

LGTM 👍

PR #51, opened 2025-05-28
fix(examples): update examples and its data

mistrykaran91:

### DIFF
--- a/src/app/examples/si-navbar/si-navbar-primary-custom-content.html
+++ b/src/app/examples/si-navbar/si-navbar-primary-custom-content.html
@@ -4,7 +4,7 @@
   [primaryItems]="primaryItems"
   [account]="{
     title: 'Jane Smith',
-    email: 'jane.smith.simpl@siemens.com',
+    email: 'jane.smith.element@siemens.com',
     company: 'Siemens',
     role: 'Admin'
   }"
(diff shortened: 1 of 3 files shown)


projects/element-ng/icon/si-icon-next.component.ts
```diff
@@ -91,7 +91,7 @@ export class SiIconNextComponent {
    *
    * @example
    * ```ts
-   * import { elementUser } from '@simpl/element-icons/ionic';
+   * import { addIcons, elementUser } from '@siemens/element-ng/icon'
```
spike-rabbit, line 94: those imports are fine here. I was wrong the first time. One should not use the siemens icon thing here.

We actuall have to find a long term solution.

src/app/examples/si-navbar/si-navbar-primary-custom-content.html
```diff
@@ -4,7 +4,7 @@
   [primaryItems]="primaryItems"
   [account]="{
     title: 'Jane Smith',
-    email: 'jane.smith.simpl@siemens.com',
+    email: 'jane.smith@siemens.com',
```
spike-rabbit, line 7: We still should have prefix here:
```suggestion
    email: 'jane.smith.element@siemens.com',
```

spike-rabbit (changes requested):
Thanks, just small stuff

PR #72, opened 2025-06-03
test(e2e): optional screenshot delay after hiding .e2e-ignore

dr-itz:
When an example uses full height, hiding the e2e control panel will cause a change in layout that can result in calculated changes that are not immediately rendered. E.g. the vertical wizard.

---


timowolf (approved):
Many thanks. This looks very good to me.

PR #104, opened 2025-06-17
feat(tabs-next): add si-tabs-next

chintankavathia:

---

### DIFF
--- a/projects/element-ng/tabs-next/si-tab-next.component.ts
+++ b/projects/element-ng/tabs-next/si-tab-next.component.ts
@@ -1,0 +1,70 @@
+/**
+ * Copyright Siemens 2016 - 2025.
+ * SPDX-License-Identifier: MIT
+ */
+import { NgClass } from '@angular/common';
+import { ChangeDetectionStrategy, Component, model, OnDestroy } from '@angular/core';
+import { SiIconNextComponent } from '@siemens/element-ng/icon';
+import { SiTranslateModule } from '@siemens/element-translate-ng/translate';
+
+import { SiTabNextBaseDirective } from './si-tab-next-base.directive';
+
+/** @experimental */
+@Component({
+  selector: 'si-tab-next',
+  templateUrl: './si-tab-next.component.html',
+  styleUrl: './si-tab-next.component.scss',
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  imports: [NgClass, SiIconNextComponent, SiTranslateModule],
+  host: {
+    '[class.active]': 'active()',
+    '[attr.aria-selected]': 'active()',
+    '(click)': 'selectTab()',
+    '(keydown.enter)': 'selectTab()'
+  }
+})
+export class SiTabNextComponent extends SiTabNextBaseDirective implements OnDestroy {
+  /**
+   * Whether the tab is active or not.
+   * If set to `true`, the tab will be selected and its content will be displayed.
+   * @defaultValue false
+   * */
+  override readonly active = model(false);
+
+  /** @internal */
+  selectTab(retainFocus?: boolean): void {
+    const tabs = this.tabset.tabPanels();
+    const newTabIndex = this.index();
+    const currentTabIndex = this.tabset.activeTabIndex();
+    let continueWithSelection = newTabIndex !== currentTabIndex;
+
+    if (continueWithSelection && currentTabIndex !== -1) {
+      const currentTab = tabs[currentTabIndex];
+      const deselectEvent = {
+        target: currentTab,
+        tabIndex: currentTabIndex,
+        cancel: () => {
+          continueWithSelection = false;
+          currentTab.active.set(true);
+        }
+      };
+
+      currentTab.active.set(false);
+      this.tabset.deselect.emit(deselectEvent);
+    }
+
+    if (continueWithSelection) {
+      this.active.set(true);
+      this.tabset.focusKeyManager?.updateActiveItem(newTabIndex);
+    }
+    if (!this.isTabButtonFullyVisible()) {
+      this.tabset.updateVisibleTabIndexes(newTabIndex, 'next');
+      this.tabset.focusKeyManager?.updateActiveItem(newTabIndex);
+    }
+    if (retainFocus) {
+      setTimeout(() => {
+        this.tabset.focusKeyManager?.activeItem?.focus();
+      });
+    }
+  }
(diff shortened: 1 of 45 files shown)


projects/element-ng/tabs-next/si-tab-next-base.component.ts
```diff
@@ -0,0 +1,183 @@
...
+  selector: 'si-tab-next-base',
+  templateUrl: './si-tab-next.component.html',
+  styleUrl: './si-tab-next.component.scss',
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  imports: [NgClass, SiIconNextComponent, SiTranslateModule],
+  host: {
+    class: 'nav-link focus-inside px-5 si-title-1',
+    role: 'tab',
+    '[class.disabled]': 'disabledTab()',
+    '[class.hidden]':
```
spike-rabbit, line 36: In what case do we need hide tabs?

projects/element-ng/tabs-next/si-tab-next-base.component.ts
```diff
@@ -0,0 +1,183 @@
...
+   *
+   * @defaultValue false
+   */
+  readonly closable = input(false, {
+    transform: booleanAttribute
+  });
+  /** Event emitter to notify when a tab is closed. */
+  readonly closeTriggered = output();
+
+  readonly badgeIsNumber = computed(() => {
```
spike-rabbit, line 89: can be protected or internal

projects/element-ng/tabs-next/si-tab-next-base.component.ts
```diff
@@ -0,0 +1,183 @@
...
+    return typeof this.badgeContent !== 'boolean';
+  });
+
+  readonly tabButton = inject<ElementRef<HTMLElement>>(ElementRef);
+  readonly tabContent = viewChild('tabContent', { read: TemplateRef });
```
spike-rabbit, line 94: internal or protected

projects/element-ng/tabs-next/si-tab-next-base.component.ts
```diff
@@ -0,0 +1,183 @@
...
+
+  private static tabCounter = 0;
+  private retainFocus = false;
+  private indexBeforeClose = -1;
+
+  tabId = `${SiTabNextBaseComponent.tabCounter++}`;
```
spike-rabbit, line 100: protected or alike

projects/element-ng/tabs-next/si-tab-next-base.component.ts
```diff
@@ -0,0 +1,183 @@
...
+  protected readonly icons = addIcons({ elementCancel });
+  protected tabset = inject(SI_TABSET_NEXT);
+  readonly index = computed(() =>
```
spike-rabbit, line 103: protected or alike

projects/element-ng/tabs-next/si-tab-next.component.ts
```diff
@@ -0,0 +1,74 @@
...
+  extends SiTabNextBaseComponent<SiTabNextComponent>
+  implements OnDestroy
+{
+  /**
+   * Whether the tab is active or not.
+   * If set to `true`, the tab will be selected and its content will be displayed.
+   * @defaultValue false
+   * */
+  override readonly active = model<boolean>(false);
+  override activateTab(): void {
```
spike-rabbit, line 35: Can we drop this methode and just make active.set(true);

I guess you can start using effect and linkedSignal since we now know that they are unchanged in v20

projects/element-ng/tabs-next/si-tabset-next.component.html
```diff
@@ -0,0 +1,108 @@
+<div class="tabs-container">
+  <div
+    #tabContainer
+    class="tab-container-buttonbar-list nav nav-tabs"
+    (siResizeObserver)="resize($event)"
+    (wheel)="mouseScroll($event)"
```
spike-rabbit, line 6: do we still need this? Since we using browser scroll I thought this would also be gone

src/app/examples/si-tabs/si-tabs-next-routing.html
```diff
@@ -0,0 +1,7 @@
+<si-tabset-next>
+  <a si-tab-next heading="Home" aria-label="Home" [routerLink]="tabLink1"></a>
```
spike-rabbit, line 2: I guess you can drop the aria-label here. I guess it is just the linter complaining. Maybe adjust the aria rule in config. We already did this for something else I guess.

I will check if we may can bring those exceptions to simpl/lint

spike-rabbit (changes requested):
It works really well now. Just a few technical remarks:

1. Why do you use index to identify tabs. I guess just using the actual reference should also be fine
2. Please check all properties / functions and make as restrictve as possible

projects/element-ng/tabs-next/si-tab-next-base.component.ts
chintankavathia: removed

chintankavathia:
> It works really well now. Just a few technical remarks:
> 
> 1. Why do you use index to identify tabs. I guess just using the actual reference should also be fine
> 2. Please check all properties / functions and make as restrictve as possible

index is used because we have use cases where we need to get next or previous tab in certain actions like delete, using keyboard to move between tabs having disabled tab in between and so on.

projects/element-ng/tabs-next/si-tab-next.component.ts
```diff
@@ -0,0 +1,70 @@
...
+    '(keydown.enter)': 'selectTab()'
+  }
+})
+export class SiTabNextComponent extends SiTabNextBaseDirective implements OnDestroy {
+  /**
+   * Whether the tab is active or not.
+   * If set to `true`, the tab will be selected and its content will be displayed.
+   * @defaultValue false
+   * */
+  override readonly active = model.required<boolean>();
```
spike-rabbit, line 32: sorry, @chintankavathia I was failing with my last suggestion. No need for required here:
```suggestion
  override readonly active = model(false);
```

if this blocks the usage undefined, than this:
```ts
  override readonly active = model<boolean>(false);
```

spike-rabbit (changes requested):
Just one little thing I suggested wrong the last

spike-rabbit (approved):
:+1:  THX a lot. this works amazingly.

PR #134, opened 2025-06-23
fix(si-split): round corners

Killusions:

---

### DIFF
--- a/projects/element-ng/split/si-split-part.component.scss
+++ b/projects/element-ng/split/si-split-part.component.scss
@@ -7,6 +7,32 @@ $header-height: 40px;
   display: flex;
   flex-direction: column;
   overflow: hidden;
+  border-radius: all-variables.$card-border-radius;
+
+  // To have rounded corners but not be breaking and allow the user to change the background color
+  &.bg-base-1 {
+    background-color: transparent !important; // stylelint-disable-line declaration-no-important
+
+    .si-split-part-content {
+      background-color: all-variables.$element-base-1;
+    }
+  }
+
+  &.bg-base-0 {
+    background-color: transparent !important; // stylelint-disable-line declaration-no-important
+
+    .si-split-part-content {
+      background-color: all-variables.$element-base-0;
+    }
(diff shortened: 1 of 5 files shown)


Killusions:
So TLDR: This can not be added by the user/from outside, and we do not modify the behavior of the background color, but merely move it to the child

projects/element-ng/split/si-split-part.component.scss
```diff
@@ -7,13 +7,40 @@ $header-height: 40px;
...
+    .si-split-part-content {
+      background-color: all-variables.$element-base-1;
+    }
+  }
+
+  &.bg-base-0 {
+    background-color: transparent !important; // stylelint-disable-line declaration-no-important
+
+    .si-split-part-content {
+      background-color: all-variables.$element-ui-4;
```
dr-itz, line 25: ui-4? that's for borders. did you mean base-0

timowolf:
Let us discuss offline or even postpone the change until we refactor the component. In case the user does not use the header, he could just add a card in the content. Your proposal does for example not work if the user specify `bg-info` on the split part. You may argue that he should not, but I think the solution provided comes with a mix of responsibilities. The user is setting the split-part in a declarative way and the classes he sets are overwritten from the component.

Killusions:
> Let us discuss offline or even postpone the change until we refactor the component. In case the user does not use the header, he could just add a card in the content. Your proposal does for example not work if the user specify `bg-info` on the split part. You may argue that he should not, but I think the solution provided comes with a mix of responsibilities. The user is setting the split-part in a declarative way and the classes he sets are overwritten from the component.

Makes sens

projects/element-ng/split/si-split-part.component.scss
Killusions: Nice catch, fixed

Killusions:
@spike-rabbit Why? Did you fix it?

Killusions:
I mean we can re-use the MR, but sure

PR #143, opened 2025-06-25
feat(theme): add a .text-link class for link in text

dr-itz:
When a link is used in a paragraph of text, it needs to be distinguishable from normal text. Either a contrast of at least 3:1 regarding the text color or by other means. Since colors don't always work, make it underline.

See https://dequeuniversity.com/rules/axe/4.10/link-in-text-block

---

### DIFF
--- a/docs/components/buttons-menus/links.md
+++ b/docs/components/buttons-menus/links.md
@@ -14,6 +14,8 @@ If the action taken by the user will change or manipulate data, use a [button](b
 - Text should be consistent with the title of the intended destination.
 - Use caution with links that are several words long. It is recommended that links are long enough to be understood by the user, but short enough to prevent text wrapping.
 - Avoid the term “click here,” other links to “here,” or the web address itself. Instead, use a meaningful descriptive label for the link, and match the destination site name.
+- Links in paragraphs should always be underlined to ensure they’re easily recognized.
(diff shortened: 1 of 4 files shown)


kfenner (commented):
@dr-itz I like this, simple and handy. We should add this to the docs incl. an example. Can you add this? 🙏 

@timowolf @spike-rabbit any objections?

docs/components/buttons-menus/links.md
```diff
@@ -21,7 +21,10 @@ If the action taken by the user will change or manipulate data, use a [button](b
 
 ### States
 
-Links support the different styles: default/visited, hover, pressed, focused and disabled
+Links support the different styles: default/visited, hover, pressed, focused and disabled.
+
+Links inside a paragraph of text need the class `.link-text` to be accessible by always being
+underlined.
```
kfenner, line 27: @dr-itz This should be moved to the code section, as it talks about the implementation.

src/app/examples/links/links.html
timowolf, line 1: Please increase the height of the demo in the docs a bit so that no school is needed.

Screenshot 2025-06-26 at 09 31 11

timowolf (dismissed):
Awesome, just some minor comments.

kfenner (approved):

PR #160, opened 2025-07-01
feat(resize-observer): add emitInitial input to control initial resize emit

chintankavathia:

### DIFF
--- a/projects/element-ng/resize-observer/si-resize-observer.directive.ts
+++ b/projects/element-ng/resize-observer/si-resize-observer.directive.ts
@@ -2,7 +2,16 @@
  * Copyright Siemens 2016 - 2025.
  * SPDX-License-Identifier: MIT
  */
-import { Directive, ElementRef, inject, input, OnDestroy, OnInit, output } from '@angular/core';
+import {
+  booleanAttribute,
+  Directive,
+  ElementRef,
+  inject,
+  input,
+  OnDestroy,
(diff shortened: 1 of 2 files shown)


projects/element-ng/resize-observer/si-resize-observer.directive.ts
```diff
@@ -23,6 +23,8 @@ import { ElementDimensions, ResizeObserverService } from './resize-observer.serv
 export class SiResizeObserverDirective implements OnInit, OnDestroy {
   /** @defaultValue 100 */
   readonly resizeThrottle = input(100);
+  /** @defaultValue true */
+  readonly emitInitial = input(true);
```
spike-rabbit, line 27: Needs attribute converter and I think it makes sense to change the default with v48 to false.


So please add a TODO comment unless you see a reason why the default should be true.

spike-rabbit (changes requested):

projects/element-ng/resize-observer/si-resize-observer.directive.ts
dr-itz: @spike-rabbit Sorry, but this is just wrong! Having an attribute converter and then changing the default to false makes absolutely no sense at all. The default MUST remain true! Otherwise it will break almost all consumers. Doing this just so that boolean transform can be used is just pure evil.

PR #177, opened 2025-07-02
fix(tabs-next): drop only showing fully visible tabs in responsive mode

spike-rabbit:
@panch1739 already saw this and approved.

Previously the tab-next had logic to only show tabs fully visible tabs and shrink the size accordingly.
This was leading to the behavior that the menu button was jumping.

With this change tabs are cut off partially,
which is the common behavior seen in other implementations.

While fixing this, the size of the menu button is also corrected.

---


chintankavathia:
There seems to be a different behavior when I delete tab based on if active tab is before or after the tab being deleted.

if active tab is before then on delete of the tab it scrolls to make it visible in the view.

if active tab is after then container scrolls little bit but doesn't always make active tab visible on the view.

**- active tab is before the tab being deleted**
tabs-delete-before

**- active tab is after the tab being deleted (observe the scroll which doesn't bring active tab fully into the view)**
tabs-delete-after

chintankavathia (approved):
LGTM

PR #178, opened 2025-07-02
build: just install lfs directly

spike-rabbit:
No need for the script before. We have ubuntu where is available.

---


spike-rabbit:
Does not work

PR #206, opened 2025-07-04
docs(date-range-filter): update ux docs

panch1739:
This PR updates the ux docs date range filter, as well as correcting the text so it matches writing guidelines.

---

- [x] I confirm that this MR follows the [contribution guidelines](https://github.com/siemens/element/blob/main/CONTRIBUTING.md)


docs/components/forms-inputs/date-range-filter.md
```diff
@@ -1,11 +1,11 @@
 # Date Range Filter
```
fh1ch, line 1: @panch1739 just a quick question: It seems we've aligned everything now within the docs to titlecase (e.g. see https://github.com/siemens/element/pull/212/files#diff-dc0281bfcbdb924cc985f0ba1e6cf21e7536aec19d2eefd7e0083efffdf58bf3L1), so shouldn't we do the same here too?

```suggestion
# Date range filter
```

fh1ch (changes requested):
@panch1739 great stuff, I like it 🙇 

Just one tiny question/nit, let me know what you think 🏓

fh1ch (approved):
@panch1739 thanks a ton for the rework, all good now 🙇 

LGTM 👍

PR #221, opened 2025-07-07
docs(tabs): update ux docs

panch1739:
This PR updates the ux docs for tabs, as well as correcting the text so it matches writing guidelines.

---


panch1739:
> About the overflow icon is it 3 dots vertically or horizontal ? Bildschirmfoto 2025-07-10 um 14 42 59

@hbxes it was aways horizontal. I think in this case works nicely because is similar to an ellipsis. It also looks a bit better with the horizontal nature of the wizard

hbxes:
> > About the overflow icon is it 3 dots vertically or horizontal ? Bildschirmfoto 2025-07-10 um 14 42 59
> 
> @hbxes it was aways horizontal. I think in this case works nicely because is similar to an ellipsis. It also looks a bit better with the horizontal nature of the wizard

OK, but this is the only place where we use horizontal ones - right ?

PR #227, opened 2025-07-08
test(playwright): fix flakiness for filtered search playground and update snapshots

mistrykaran91:


spike-rabbit:
seems like this does not fix the issue: https://github.com/siemens/element/actions/runs/16138150699/job/45539184045

spike-rabbit:
Fixed in #239

PR #233, opened 2025-07-08
refactor(docs): update component tooltip

hbxes:

---


docs/components/status-notifications/images/tooltip-revaluate.png
timowolf, line 1: While I know this is about the tooltip, should we have a right alignment here of the button and the card or remove the right empty space?

timowolf (commented):
Perfect, just a minor comment on one screen. But not too important.

spike-rabbit:
Not exactly crucial, but the commit message should be different, as those changes are not a refactoring:
Something like this is more suitable: 

`docs(tooltip): align with whateever`

Same for other docs PRs that are labelled as refactoring.

spike-rabbit:
This PR only affects documentation, not code. Refactoring is dedicated for code changes of the actual package we deliver.

Updating images or MD files does not affect the code we deliver. It is only part of the documentation so the scope should be `docs`.

docs/components/status-notifications/images/tooltip-revaluate.png
panch1739: @timowolf I also dont understand what you mean 🤔 The images look ok for me

fh1ch:
> This PR only affects documentation, not code. Refactoring is dedicated for code changes of the actual package we deliver.
> 
> Updating images or MD files does not affect the code we deliver. It is only part of the documentation so the scope should be `docs`.

See also https://github.com/angular/angular/blob/main/contributing-docs/commit-message-guidelines.md#type:

- **docs:** Documentation only changes, *aka. everything related to `.md` files and referenced stuff*
- **refactor:** A code change that neither fixes a bug nor adds a feature, *aka. has to be a code change in this context*

docs/components/status-notifications/images/tooltip-revaluate.png
timowolf: This is really my mistake and an interesting difference between GitLab and GitHub. 

- Both systems show the old image on the left side and the new image on the right side
- For image review I mostly use the swipe function, but:
  - GitLab shows the old image first and when moving the handle to right the new image comes in.
  - GitHub shows the new image first and when moving the handle to the right the old image comes in.

This is confusing.... So I reviewed the wrong image :-)

All good now.

timowolf (approved):
Many thanks. All good.

PR #237, opened 2025-07-08
docs(list-details): add table resize to example

Killusions:
> Make sure the table column resize if the use moves the split.

---


Killusions:
@timowolf This is the fix to your follow-up, currently failing only because main fails

PR #247, opened 2025-07-09
docs(typeahead): update ux docs

panch1739:
This PR updates the ux docs for typeahead, as well as correcting the text so it matches writing guidelines.

---


docs/components/sorting-filtering/images/typeahead.png
hbxes, line 1: There is no focus ring anymore, is this correct ?

hbxes: No, if it's a real condition, then it's OK. I just thought because it was in the old image

fh1ch (approved):
@panch1739 same here, great work, much obliged 🙇 

LGTM 👍

PR #252, opened 2025-07-10
Feat/add about component

chintankavathia:


playwright/snapshots/static.spec.ts-snapshots/si-about--si-about-api-element-examples-chromium-light-linux.png
chintankavathia, line 1: @kfenner, @spike-rabbit  I guess the favicon is getting replaced on github at runtime and this fails the e2e. We would need some other icon image for this here.

kfenner: OK will take care of it 👍

PR #253, opened 2025-07-10
feat(translate): use provider function to create service instances

spike-rabbit:
In order to achieve 100% standalone, we need to get rid of all modules. Until now, we used a module to ensure
that a correct service instances is created.
This change adds an alternative `injectSiTranslateService`, which shall replace the `SiTranslateModule`.

Having a specialized inject function follows a pattern seen in other libraries like [angular query](https://tanstack.com/query/v5/docs/framework/angular/guides/queries).

---


spike-rabbit:
This was already done in another PR

PR #256, opened 2025-07-11
feat(translate): resolve error with new application unit test builder

spike-rabbit:
This PR fixes the issues with the new angular application unit test builder.
Both are needed as Angular fails with modules in certain cases.
Yet it also helps to fully go standalone.

A follow-up will drop all other usages of the module.
This PR only contains the necessary changes.

We need to keep the module for compatibility.

## fix(translate): prevent missing $localize errors
Moves the initialization out of the module constructor
as the constructor is not reliably called.
The new version ensures that `$localize` is patched
early enough before any other angular initializations.

## feat(use) provider function to create service instances
In order to achieve 100% standalone, we need to get rid of all modules.
Until now, we used a module to ensure
that a correct service instances is created.
This change adds an alternative `injectSiTranslateService`,
which shall replace the `SiTranslateModule`.

Having a specialized inject function follows a pattern seen
in other libraries like [angular query](https://tanstack.com/query/v5/docs/framework/angular/guides/queries).

---

### DIFF
--- a/projects/element-translate-ng/translate/si-localize.ts
+++ b/projects/element-translate-ng/translate/si-localize.ts
@@ -25,18 +25,17 @@ const $localize = (strings: TemplateStringsArray, ...expressions: string[]): Tra
 const siResolveLocalize = (key: string, value: string): TranslatableString =>
   inject(SiTranslatableService).resolveText(key, value);
 
-/**
- * Initializes the Element version of $localize. Can be called multiple times.
- */
-export const initSiLocalize = (): void => {
-  (Zone.current as any)._properties.siResolveLocalize = siResolveLocalize;
-  globalScope.$localize = $localize;
-};
+// Always register in the current zone. This is needed in MFE setups, where $localize is already patched.
+(Zone.current as any)._properties.siResolveLocalize = siResolveLocalize;
 
-// Init $localize in the current zone
+// Patch $localize in the global scope if it was not already patched by ourselves or Angular.
+// The default $localize function by Angular just throws an error.
 try {
   // eslint-disable-next-line @typescript-eslint/no-unused-expressions
   globalScope.$localize`:@@si-localize:This is a test for $localize`;
 } catch {
-  initSiLocalize();
+  globalScope.$localize = $localize;
 }
+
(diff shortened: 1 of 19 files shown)


spliffone:
@spike-rabbit Why do we still use the `SiTranslateModule`? I guess we should use instead a provider function in the standalone components.

spliffone:
The change resolve the new karma builder issue independent of the just in time compilation problem mentioned in https://github.com/angular/angular/issues/62573

projects/element-translate-ng/translate/si-localize.ts
```diff
@@ -32,3 +32,11 @@ export const initSiLocalize = (): void => {
   (Zone.current as any)._properties.siResolveLocalize = siResolveLocalize;
   globalScope.$localize = $localize;
 };
+
+// Init $localize in the current zone
```
fh1ch, line 36: WDYT if we're a bit more specific here?

```suggestion
// Init $localize in the current zone if not already available
```

fh1ch (changes requested):
@spike-rabbit really great work here, thanks a ton 🙇 

The changes all look good and even gave it a short local try. Hence only two nits/questions from my end. WDYT?

projects/element-translate-ng/translate/si-localize.ts
spike-rabbit: Slighlty had to adjust to calls, but comment is improved as well.

projects/element-translate-ng/translate/si-translate.module.ts
```diff
@@ -28,17 +26,8 @@ import { SiTranslateServiceBuilder } from './si-translate.service-builder';
...
 })
-export class SiTranslateModule {
-  constructor() {
-    const translateService = inject(SiTranslateService);
-    if (!translateService.prevent$LocalizeInit) {
-      initSiLocalize();
-    }
-  }
-}
+export class SiTranslateModule {}
```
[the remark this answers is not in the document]
spike-rabbit: yes, I will do in the other PR, where I replace the internal usage of this module.
Otherwise we have a conflict with the linter

timowolf:
Many thanks, looks really nice.

Issue #234, opened 2025-07-08
Enable new linter rules


spike-rabbit:
These rules should be enabled:
- @angular-eslint/sort-keys-in-type-decorator
- @angular-eslint/prefer-inject
- @angular-eslint/prefer-output-emitter-ref
- @angular-eslint/template/prefer-at-empty

Please enable in separate MR

PR #262, opened 2025-07-14
refactor(lint): use inject function

prasepic:
Addresses #234 

---

### DIFF
--- a/projects/element-ng/form/si-form-validation-error.service.ts
+++ b/projects/element-ng/form/si-form-validation-error.service.ts
@@ -52,6 +52,7 @@ export class SiFormValidationErrorService {
     skipSelf: true
   });
 
+  // eslint-disable-next-line @angular-eslint/prefer-inject
   constructor(private errorMapper: SiFormValidationErrorMapper) {}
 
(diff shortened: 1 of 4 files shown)


projects/element-ng/form/si-form-validation-error.service.ts
```diff
@@ -52,6 +52,7 @@ export class SiFormValidationErrorService {
     skipSelf: true
   });
 
+  // eslint-disable-next-line @angular-eslint/prefer-inject
```
spike-rabbit, line 55: Please replace ctor injection in this location and others, unless the replacement requires larger refactoring.

spike-rabbit (changes requested):

projects/element-ng/form/si-form-validation-error.service.ts
prasepic: The problem with this service is, it uses a type to construct. We cannot explicitly use inject function. We have to fallback to injection tokens (This was also the earlier approach but we removed it when we exposed the error provider)

PR #265, opened 2025-07-14
refactor(lint): prefer output instead of @Output

prasepic:
Addresses #234 

This however is a breaking change. `EventEmitter` extends `Subject` by default and is pipeable. 
`output` signal cannot be piped on.

---

### DIFF
--- a/projects/element-ng/split/si-split.component.ts
+++ b/projects/element-ng/split/si-split.component.ts
@@ -2,6 +2,7 @@
  * Copyright (c) Siemens 2016 - 2025
  * SPDX-License-Identifier: MIT
  */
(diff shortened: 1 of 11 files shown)


projects/element-ng/split/si-split.component.ts
```diff
@@ -265,6 +267,7 @@ export class SiSplitComponent implements AfterContentInit, OnChanges {
             merge(fromEvent(this.document, 'mouseup'), fromEvent(this.document, 'touchend'))
           )
         )
+        .pipe(takeUntilDestroyed(this.destroyRef))
```
prasepic, line 270: This is needed due to https://github.com/angular/angular/issues/60110 .

They've made this a warning in 19.2.x

PR #270, opened 2025-07-14
fix(charts): empty series name in tooltip

akashsonune:
The series name in the chart tooltip is empty

image

This is because the `series.name` can be an empty string, in which case, it should fallback to the `series.seriesName` . this does not work right now because of how the `??` operator works
 

### DIFF
--- a/projects/charts-ng/src/components/si-chart-cartesian/si-chart-cartesian.component.ts
+++ b/projects/charts-ng/src/components/si-chart-cartesian/si-chart-cartesian.component.ts
@@ -69,7 +69,8 @@ export class SiChartCartesianComponent extends SiChartComponent implements OnCha
       const name = isPie
         ? series.data.name
         : useName
-          ? (series.name ?? series.seriesName)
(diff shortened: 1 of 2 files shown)


spike-rabbit:
@akashsonune please describe why this is needed.

akashsonune:
> @akashsonune please describe why this is needed.

Added in description

akashsonune:
> Thx, What is the reason why someone provides an empty string instead of undefined?

In the tooltip formatter, `name` field stands for data value which is usually specified in case of category type of axis. So in other cases, this comes out to be an empty string. this is sent from echarts and we have no control over it.
https://echarts.apache.org/en/option.html#series-line.tooltip.formatter

dr-itz:
> Ok, makes sense for me. Please add the ignore comments. Seems like strings do not account as primitive

String counts as a primitive, but here we have `any` (from the echarts callback), so the rule still complains.

akashsonune:
Ok, what do we do then? Ignore the rule here or reconsider if we want to keep the rule

spike-rabbit (approved):

PR #299, opened 2025-07-16
Fix/theme base 4 correct

fh1ch:
This PR:

- Bumps the `@simpl/brand` package to include the missing CSS variable for `base-4`.
- Adds the missing `base-4` to the CSS utility classes, so it can be used like the other `base-*` counterparts.

---


timowolf:
Many thanks.

PR #313, opened 2025-07-17
Docs/provide element logo and favicon

kfenner:
This PR has two commits:

1. Provide Element logo variants in an "artwork" folder
2. Replace the SiMPL favicon with a new Element favicon as default (is replaced with Siemens favicon in the docs pipeline)

Question for commit 2: Why do we have the favicon file so many times? Could we bring this down to one and if needed copy the file? Can somebody help with this? 

Btw: I went for the size of 96x96px for the favicon.png - before it was inconsistent with two variants, not sure what is really best and if it matters much these days.

/cc @mistrykaran91 

---


fh1ch (approved):
@kfenner as discussed offline with @spike-rabbit: Lovely work, thanks a lot 🙇 

Teh changes look good and it makes sense to publish all the official assets, hence no feedback from my end.

LGTM 👍

spike-rabbit (approved):

PR #316, opened 2025-07-17
fix(select): do not highlight option as active by default

chintankavathia:
select options should show active background only after user use keyboard arrows to navigate between options


chintankavathia:
As discussed we will keep behavior as is. closing this one.

PR #317, opened 2025-07-18
feat(wizard): switch to footer navigation as default

chintankavathia:
BREAKING CHANGE:
`inlineNavigation` is now default set to `false` and will render navigation buttons in footer which is recommended. You can switch back to inline by explicitly setting `inlineNavigation` to `true`.


spike-rabbit:
Please update the commit message a little:

---

BREAKING CHANGE:
The `si-wizard` now has the navigation buttons by default in the footer.

To restore the old behavior set `SiWizardComponent.inlineNavigation` to `true`:
```html
<si-wizard inlineNavigation>
   ...
</si-wizard>
```

spike-rabbit (approved):
:+1:

PR #318, opened 2025-07-18
fix(wizard): update disabled styles to match with figma

chintankavathia:


spike-rabbit (changes requested):
Can you please add a vrt that covers this case

PR #323, opened 2025-07-18
Docs/add missing example 2

spike-rabbit:
None of those example has VRT.

---


spike-rabbit:
I would actually be in favor of dropping `si-` everywhere.
Let's discuss this after n20 update, which also has the new styleguide

kfenner (approved):
LGTM 👍

PR #359, opened 2025-07-22
build(release): update version in dependent package right after setting the version

spike-rabbit:
This should ideally solve our release problem.

The Problem is, that when bumping a package inside the workspace, the dependent packages are not automatically updated.
So they have an invalid peer dependency, since they depend on the old version.
This is hopefully solved by updating all dependent packages directly after a package was bumped.
We previously updated all packages after all were bumped.

This occurs since adding dashboards-ng.
We previously did not have this problem, as I forgot adding `element-ng` to the workspaces.

---

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,6 +9,7 @@
       "version": "47.7.0",
       "license": "MIT",
       "workspaces": [
+        "projects/element-ng",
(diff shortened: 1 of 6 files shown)


fh1ch (approved):
LGTM 👍

PR #367, opened 2025-07-22
feat(icon): replace current `si-icon` with `si-icon-next`

spike-rabbit:
Dropping the current `si-icon` and renaming `si-icon-next` to `si-icon`.

BREAKING CHANGE: The `si-icon` component was rewritten leading to a few API changes.
The overall goal of this change is to ensure a similar behavior compared to
the direct use via css classes.

The most notable changes are:
- dropping the default size
- dropping most inputs in favor of using css classes

Single colored icon can be converted like this:

```html
<!-- before -->
<si-icon icon="element-user" color="text-danger" />
<!-- after -->
<si-icon icon="element-user" class="icon text-danger" />
```

**Important:** Previously, the class `icon` was automatically applied. Unless not needed,
it must now be applied manually.
The icon class scales up the icon compared to its surrounding text.

Stacked icons need to be constructed in HTML directly.
If applicable, the `si-status-icon` component should be used instead.

```html
<!-- before -->
<si-icon
  icon="element-circle-filled"
  color="status-success"
  stackedIcon="element-state-tick"
  stackedColor="status-success-contrast"
  alt="Success"
/>

<!-- after -->
<div class="icon-stack icon" aria-label="Success">
  <si-icon icon="element-circle-filled" class="status-success" />
  <si-icon icon="element-state-tick" class="status-success-contrast" />
</div>
```
---

- [X] I confirm that this MR follows the [contribution guidelines](https://github.com/siemens/element/blob/main/CONTRIBUTING.md).


timowolf (changes requested):
- Keep old icon as legacy
- Change breaking change note as the icon class sets 1.5rem and does not scale with fonts

timowolf (approved):
LGTM

PR #370, opened 2025-07-23
refactor(tabs-next): use the base class where possible

spike-rabbit:
Previously, in theory the order of tabs could have been wrong. This should never be an issue though, as mixing routing and the buttons tabs should e avoided.

Anyway, this is fixed with this.

---


kfenner (approved):
LGTM 👍

PR #376, opened 2025-07-23
fix(theme): correct typography definition for H1

fh1ch:
Adapts the H1 typography definition to be in-sync with the Figma specification.

---


timowolf:
As discussed, we should fix them all and align the font sizes, including h4, h5, ... with the introduction of the new fonts. This will be a big pixel update anyway. So we should take this change and the breaking change phase now to include them all.

panch1739:
> As discussed, we should fix them all and align the font sizes, including h4, h5, ... with the introduction of the new fonts. This will be a big pixel update anyway. So we should take this change and the breaking change phase now to include them all.

@timowolf yeah i agree, but i also we run into a conflict between the semantic heading levels (like H1, H2, etc.) and the visual styles defined.

In our system, H1 is designed to be very large, visually meant for a top-level, prominent use case (like a login page). But when we try to use H1 semantically in a regular application page (e.g. a page header), that size feels out of place: too big or overpowering for the layout.

So we are stuck between wanting to follow semantic HTML rules for accessibility and structure (using one H1 per page),
but also apply a more appropriate visual size, which may actually look more like our H2.

So... I either break semantic structure to get the right visual result, or I use proper semantics but the UI looks off. I’m not sure how to reconcile this in a consistent and scalable way yet. And this is a valid concern from @kfenner

kfenner:
@panch1739 Thanks for bringing up my concern. I'm fine with fixing stuff or documenting/introducing things we are lacking. But I'd put the change of semantics on hold before we have a fulll internal alignment.

PR #379, opened 2025-07-23
feat(tabs): align with UX specs

spike-rabbit:
- vertical center badges
- drop min-width in icon only mode
- reduce spacing around close button
- less padding-end if close button

---


panch1739:
@spike-rabbit much better!! But I have more points

1. Does it make sense to have a min width of 100px for the icon only version? To me, it feels off. The purpose of setting a min width is to ensure that very short text doesn’t break the layout, but in the icon-only case, all icons are the same size.

Since the whole point of an icon only tab is to enable a more compact, dense layout, I’d suggest removing the min width for this variant:

image

2. The distance between the close button and the text feels a bit too much now. I propose adjusting it to 4px:

image

3. I noticed that in the closable version, the left and right paddings are different to compensate for the close button. I know you hate this, but it does look better:

image

image


One more thing...the badge was never meant to support text + icon. UX defines that the badges should be `text-only` or `icon only`... I think we shouldn't prevent this from the code, as it may be used by projects already, but i would remove the icon from Lobby and Pantry from the demo page, to avoid advertising an usage that we don't prefer :

Screenshot 2025-07-23 at 16 38 28


@fh1ch since you are a clear core member of this team without any other work assigned, what do you think? Do you think is too many changes without a proper discussion?

fh1ch:
> clear core member of this team without any other work assigned, what do you think? Do you think is too many changes without a proper discussion?

@panch1739 really good points, thanks a lot for the in-depth review. First things first, I do agree with your points.

1. Yes, having a min-wdith for icon-only use-cases looks off and is also no in-line with other components that support icon only.
2. You're right, I like the proposed spacing also better.
3. Agree, but I frankly can't comment on the implementation effort here, I leave this up to @spike-rabbit if this is feasible to do as part of this change or if we should do a follow-up issue.
4. True, that should be a really easy one to do. Considering that we heavily use declarative interfaces and hence allow heavy customisation, we shouldn't prevent it in code, but should surely **not** promote bad examples in our demo.

No, I frankly don't see an issue with doing all of the changes in this PR, as I'd consider them detail enhancements and not general changes of our design system.

spike-rabbit:
@panch1739 please check again. Adjusted all your points

panch1739:
> @panch1739 please check again. Adjusted all your points

@spike-rabbit love itt, much better, what do you think?

panch1739:
> > clear core member of this team without any other work assigned, what do you think? Do you think is too many changes without a proper discussion?
> 
> @panch1739 really good points, thanks a lot for the in-depth review. First things first, I do agree with your points.
> 
> 1. Yes, having a min-wdith for icon-only use-cases looks off and is also no in-line with other components that support icon only.
> 2. You're right, I like the proposed spacing also better.
> 3. Agree, but I frankly can't comment on the implementation effort here, I leave this up to @spike-rabbit if this is feasible to do as part of this change or if we should do a follow-up issue.
> 4. True, that should be a really easy one to do. Considering that we heavily use declarative interfaces and hence allow heavy customisation, we shouldn't prevent it in code, but should surely **not** promote bad examples in our demo.
> 
> No, I frankly don't see an issue with doing all of the changes in this PR, as I'd consider them detail enhancements and not general changes of our design system.

@fh1ch awesomeeeee! thank you so much!

PR #390, opened 2025-07-24
refactor(autocomplete): use signal queries

chintankavathia:

---

### DIFF
--- a/projects/element-ng/autocomplete/si-autocomplete-listbox.directive.ts
+++ b/projects/element-ng/autocomplete/si-autocomplete-listbox.directive.ts
@@ -4,20 +4,17 @@
  */
 import { ActiveDescendantKeyManager } from '@angular/cdk/a11y';
 import {
-  AfterContentInit,
   ChangeDetectorRef,
-  ContentChildren,
   DestroyRef,
   Directive,
   inject,
   input,
-  OnChanges,
   OnInit,
   output,
-  QueryList,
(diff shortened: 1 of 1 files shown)


projects/element-ng/autocomplete/si-autocomplete-listbox.directive.ts
```diff
@@ -115,6 +108,9 @@ export class SiAutocompleteListboxDirective<T> implements OnInit, OnChanges, Aft
   }
 
   get active(): SiAutocompleteOptionDirective<T> | null {
-    return this.keyManager?.activeItem ?? null;
+    if (this.keyManager.activeItemIndex === null) {
+      return null;
+    }
+    return this.options().at(this.keyManager.activeItemIndex) ?? null;
```
spike-rabbit, line 114: Why this change? 

`activeItem` should still be working

spike-rabbit (changes requested):

projects/element-ng/autocomplete/si-autocomplete-listbox.directive.ts
chintankavathia: when options get updated dynamically the `keyManager.activeItem` still refs to the old option item. could be bug in cdk as activeItemIndex is updated correctly.

spike-rabbit (approved):
:+1:

PR #394, opened 2025-07-24
chore: add issue templates

ljanner:
Adds issue templates for bug reports and feature requests inspired by common practices and our friends over at https://github.com/siemens/ix

Blank issues are enabled by default and I did not see a need to add a `config.yml` inside `.github/ISSUE_TEMPLATE/` as we don't have public communication channels yet.

Both templates use the label _**triage**_ that does not exist yet and would have to be created. This is just a proposal, newly created issues could also be labeled directly with **bug** and **feature** accordingly.

---

### DIFF
--- a/.github/ISSUE_TEMPLATE/feature_request.yml
+++ b/.github/ISSUE_TEMPLATE/feature_request.yml
@@ -1,0 +1,23 @@
+name: Feature request
+description: Suggest a feature for Siemens Element
+labels: ['triage']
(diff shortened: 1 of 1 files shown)


timowolf:
Sounds good to me. I think we can start like that. We have also many feature requests that are component related, but let us see how it works out.

timowolf (approved):
Awesome, many thanks.

PR #398, opened 2025-07-25
feat(switch): align switch label padding with checkboxes

spliffone:

- align label spacing to checkbox with 8px space between switch and label

---


spike-rabbit:
This is already a breaking change, since one can no longer combine `.col` with `.form-switch` on the same element.
Please only fix the spacing even though will rewrite that line again later. But at least we have already updated screenshots that way.

Style is not the correct scope btw, please have a look at the contributing. (style: Changes that do not affect the meaning of the code (white-space, formatting, missing semi-colons, etc)). I think a fix is correct in this case.

spliffone:
> This is already a breaking change, since one can no longer combine `.col` with `.form-switch` on the same element. Please only fix the spacing even though will rewrite that line again later. But at least we have already updated screenshots that way.
> 
> Style is not the correct scope btw, please have a look at the contributing. (style: Changes that do not affect the meaning of the code (white-space, formatting, missing semi-colons, etc)). I think a fix is correct in this case.

The commit now only contain the spacing without breaking

src/app/examples/si-charts/interactive/interactive.html
```diff
@@ -67,55 +67,63 @@
...
-            <input
-              class="form-check-input col-sm-2"
-              type="checkbox"
-              id="axisPointer"
-              role="switch"
-              [checked]="axisPointer"
-              (change)="toggleAxisPointer()"
-            />
-            <label class="form-check-label col-sm-8" for="axisPointer">Axis Pointer</label>
+          <div class="col">
```
spike-rabbit, line 70: can you please also move those changes out of this PR as they are not yet needed

spike-rabbit (changes requested):
Please also adjust the commit message. `style` is not the correct type

spike-rabbit (approved):
:+1:

PR #407, opened 2025-07-27
feat(card): add support for sub-heading to cards

fh1ch:
Implements support for card sub-headings within the [si-card](https://element.siemens.io/components/layout-navigation/cards/) component based on @panch1739's UX design as well as Bootstrap's example: https://getbootstrap.com/docs/5.0/components/card/#titles-text-and-links.

---


projects/element-ng/card/si-card.component.html
```diff
@@ -12,9 +12,16 @@
   @if (heading() || displayContentActionBar()) {
     <div class="card-header d-flex justify-content-between">
       <ng-content #cardHeaderIcon select="[headerIcon]" />
-      @if (heading()) {
-        <div class="text-truncate">{{ heading() | translate }}</div>
-      }
+      <div>
+        @if (heading()) {
+          <div class="text-truncate si-title-2">{{ heading() | translate }}</div>
```
fh1ch, line 17: @spike-rabbit another quick design system topic: If we don't set `si-title-2` here, we'll get a wrong default line-height hence cutting off parts of the text. While this would be correct from a design system perspective, it causes quite a lot of VRT changes.

Alternatively, we could also extend the line-height overwrite that we had in place previously, but no longer applies due to the nesting:

```diff
diff --git a/projects/element-ng/card/si-card.component.html b/projects/element-ng/card/si-card.component.html
index 47b1ad7..d4447af 100644
--- a/projects/element-ng/card/si-card.component.html
+++ b/projects/element-ng/card/si-card.component.html
@@ -14,7 +14,7 @@
       <ng-content #cardHeaderIcon select="[headerIcon]" />
       <div>
         @if (heading()) {
-          <div class="text-truncate si-title-2">{{ heading() | translate }}</div>
+          <div class="text-truncate card-title">{{ heading() | translate }}</div>
         }
         @if (subHeading()) {
           <div class="text-truncate si-body-2 text-secondary pt-2">{{
diff --git a/projects/element-theme/src/styles/bootstrap/_card.scss b/projects/element-theme/src/styles/bootstrap/_card.scss
index 1b4e717..d6f218b 100644
--- a/projects/element-theme/src/styles/bootstrap/_card.scss
+++ b/projects/element-theme/src/styles/bootstrap/_card.scss
@@ -90,6 +90,7 @@
   background-color: variables.$card-cap-bg;

   &,
+  .card-title,
   .icon {
     line-height: typography.$si-font-size-title-2;
   }
```

WDYT?

PR #410, opened 2025-07-28
feat(accordion): allow using hcollapsed as html attribute

chintankavathia:
allows to use `hcollapsed` as html element attribute.

For example, `<si-accordion [hcollapsed]="true" />` can now be also be written like `<si-accordion hcollapsed />`


spike-rabbit:
I think theoretically hcollapsed should actually be a model as the "real" state changes by user interaction. So I think not having the converter here allows us to maybe change this later.

PR #412, opened 2025-07-28
fix(date-range-filter): allow empty reference point

spliffone:
**DateRangeFilter issue:**

1. Open DateRangeFilter example 
2. Switch on advanced mode
3. Clear both `Reference point` and `Range`
4. Switch off advanced mode

Results in an exception and the datepicker isn't shown.

---

### DIFF
--- a/projects/element-ng/date-range-filter/si-date-range-filter.component.html
+++ b/projects/element-ng/date-range-filter/si-date-range-filter.component.html
@@ -63,7 +63,7 @@
   } @else {
     <si-datepicker
       [config]="dateRangeConfig()"
-      [focusedDate]="dateRange.end ?? dateRange.start"
+      [focusedDate]="focusedDate()"
       [(dateRange)]="dateRange"
       (dateRangeChange)="updateFromDateRange()"
     />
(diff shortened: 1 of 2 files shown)


dr-itz:
The other thing to mention here: the datepicker should allow undefined `focusedDate` and do the `today()` thing on its own instead of crashing

spliffone:
> Now I took a close look to the issue:
> 
> > The other thing to mention here: the datepicker should allow undefined `focusedDate` and do the `today()` thing on its own instead of crashing
> 
> Yes supporting undefined in general makes sense. But we put us in a bad spot with the current API. What the datepicker needs is an input, to set the `initalFocusedDate` which should allow (and support) undefined and then default it to something meaningful.
> 
> The actual `focusedDate` must always be defined, as we always need it to show the calendar. In my opinion, this property should have never been part of the public API. The type is also wrong, as it allows undefined, yet undefined causes an error.
> 
> But before we start changing API stuff, I will just merge here.

The issue is not that we we have a focusDate, also setting undefined works like expected which automatically result in today. The issue was that an invalid date was passed which cause the stacktrace.

PR #415, opened 2025-07-28
fix(tabs-next): ensure that the active tab is focussed by default

spike-rabbit:
Unless actively operated,
the focuskeymanger should initially always point to the active tab. Previously, this only worked if the tabset is initialized with tabs on load. If tabs are added later or the active flag was changed programmatically, tha focuskeymanager was not updated.
This commit ensures
that the focuskeymanager is always updated if the active flag has changed.

---


fh1ch (approved):
@spike-rabbit thanks a lot, all good here 🙇 

LGTM 👍

PR #418, opened 2025-07-28
refactor(tabs-next): fully rely on routing for link tabs

spike-rabbit:
Previously, the active flag of all tabs implementation must have been a WritableSignal.
This was causing problems in the link tabs, as there everything is managed by the router.
So we needed an effect to react to changes on the active signal.

Wit this change, we now have methods to programmatically select and deselect tabs.
Those methods never cause the emitting of events and could also be used by applications.
The active flag is now also readonly in the base class to prevent setting it directly
if the router is used.

---

### DIFF
--- a/projects/element-ng/tabs-next/si-tab-next-link.component.ts
+++ b/projects/element-ng/tabs-next/si-tab-next-link.component.ts
@@ -2,11 +2,12 @@
  * Copyright (c) Siemens 2016 - 2025
  * SPDX-License-Identifier: MIT
  */
-import { ChangeDetectionStrategy, Component, effect, inject, signal } from '@angular/core';
-import { takeUntilDestroyed } from '@angular/core/rxjs-interop';
+import { ChangeDetectionStrategy, Component, inject } from '@angular/core';
+import { toSignal } from '@angular/core/rxjs-interop';
 import { Router, RouterLink, RouterLinkActive } from '@angular/router';
 import { SiIconNextComponent } from '@siemens/element-ng/icon';
 import { SiTranslatePipe } from '@siemens/element-translate-ng/translate';
+import { startWith } from 'rxjs/operators';
 
 import { SiTabBadgeComponent } from './si-tab-badge.component';
 import { SiTabNextBaseDirective } from './si-tab-next-base.directive';
@@ -20,10 +21,6 @@ import { SiTabNextBaseDirective } from './si-tab-next-base.directive';
   styleUrl: './si-tab-next.component.scss',
   providers: [{ provide: SiTabNextBaseDirective, useExisting: SiTabNextLinkComponent }],
   changeDetection: ChangeDetectionStrategy.OnPush,
-  host: {
-    '[class.active]': 'routerLinkActive.isActive',
-    '[attr.aria-selected]': 'routerLinkActive.isActive'
-  },
   hostDirectives: [
     {
       directive: RouterLinkActive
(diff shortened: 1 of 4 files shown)


projects/element-ng/tabs-next/si-tab-next.component.ts
```diff
@@ -32,31 +30,41 @@ export class SiTabNextComponent extends SiTabNextBaseDirective implements OnDest
...
+    const tabs = this.tabset.tabPanels();
+    const currentTabIndex = this.tabset.activeTabIndex();
+    const currentTab = tabs[currentTabIndex];
+
+    currentTab?.deSelectTab();
+    this.active.set(true);
+    super.selectTab(retainFocus);
+  }
+
+  override deSelectTab(): void {
```
fh1ch, line 67: Shouldn't this one be documented as `@ internal` or is it intended to be used outside of our code-base?

projects/element-ng/tabs-next/si-tab-next-link.component.ts
```diff
@@ -20,35 +21,31 @@ import { SiTabNextBaseDirective } from './si-tab-next-base.directive';
...
   ]
 })
 export class SiTabNextLinkComponent extends SiTabNextBaseDirective {
-  private router = inject(Router, { optional: true });
-  /** @defaultValue false */
-  override readonly active = signal(false);
+  private router = inject(Router);
   /** @internal */
-  routerLink = inject(RouterLink, { optional: true, self: true });
+  routerLink = inject(RouterLink, { self: true });
```
fh1ch, line 33: Doesn't the removal of the `optional` pose a breaking change? The tabs used to be usable without hard dependency on the Angular router, e.g. for minimal use-cases where applications were not using a router at all.

fh1ch (commented):
@spike-rabbit thanks a ton for the PR, much appreciated 💯 

Just two questions, otherwise we're good to go here 🏓

projects/element-ng/tabs-next/si-tab-next.component.ts
spike-rabbit: I did not make my mind up about this fully. Select tab should be public in my opinion. So it would be logic to keep deSelectTab as well in the public.

I am also planning to write documentation/js-docs in a follow-up PR. I guess then I also need to decide on this.

projects/element-ng/tabs-next/si-tab-next-link.component.ts
spike-rabbit: This is only the si-tab-next-link, which is designed to be used with the router. The "normal" `si-tab-next` still works without the router.

projects/element-ng/tabs-next/si-tab-next.component.ts
fh1ch: @spike-rabbit well, as long as you take care of the follow-up MR before v48, it's all fine with me. We should just avoid making such things public and then internal again after a few releases, if people start using it wrongly.

projects/element-ng/tabs-next/si-tab-next-link.component.ts
fh1ch: @spike-rabbit yep, this sounds all logical, but technically seen, it's still a breaking change. But considering that the component wasn't released yet and indeed is called `link`, I'm perfectly fine to violate semantics in this case.

fh1ch (approved):
LGTM 👍

PR #424, opened 2025-07-29
fix(dashboards-ng): cannot add or edit webcomponent widgets

chintankavathia:
fix issue where host application cannot resolve translation service when widget is added as webcomponent.


chintankavathia:
should be fixed with introduction of `t` in `element-translate-ng`

PR #438, opened 2025-07-30
fix(badge): align default text with design specs

### LINKED ISSUE #425: <badges/badges>: Change text style to body in default style
### Summary

I realised that the implementation is not matching the design for the default badge style:

Image

### What is the expected correct design?

The badges that are in default style should be using body-2 and only the emphasis variant should use title-2 style:

Image

The specs can be found in Figma

ljanner:
Previously the badge text was always using the emphasis style

Resolves #425

---


spike-rabbit:
@panch1739 please confirm that the badges now look as expected

panch1739:
> @panch1739 please confirm that the badges now look as expected

Perfect @spike-rabbit ! Thank you so much 🙏

spike-rabbit (approved):

Issue #436, opened 2025-07-30
translation layer issues


spike-rabbit:
Our translation layer is currently build in a very hacky manner to support all the features we need.
It basically consists of two parts:

1. the `$localize` function in the global scope to allow using with `@angular/localize`
2. the `SiTranslatePipe/Service` to connect with `ngx-translate`

The `$localize` function is the problematic part as we are facing some constraints here which are hard to solve all at once:
1. It must be in the global scope --> only one instance, independent of how many angulars apps are running in parallel
2. it uses injectionContext --> it must switch between different angular inject functions from different bundles (we use zone.js to detect the angular bundle)
3. it must be patched in the global scope before being called. This means before a component is instantiated
4. it must be patched in the global scope after we are sure, that not `@angular/localize` i used

We previously managed to cover all those requirements by using ng-modules, since a module constructor is always evaluated on import in a component. Since angular broke ng-modules in certain cases, we can no longer do this.

Our first fix approach was to patch it when the translation layer is imported. This works except, when there are multiple angular instances, but they share the element-translate-ng package. Then `$localize` is not updated for the subsequent angular instances.

Based on this I see now two major ways to solve this problem:

# Option 1: Wrapping of `$localize` calls

This is an alternative we already considered before, which would require us to transform every usage of `$localize`.
Usage wise, it could look like this:

```ts
import { t } from `@siemens/element-translate-ng`

class SiElementComponent {
  readonly before = input($localize`:@@key:value`);
  readonly after = input(t(() => $localize`:@@key:value`));
}
```

The relevant change here, that we now control when `$localize` is evaluated and that `t` is a normal import, so we can use `inject` inside without the need of looking something up in `zone.js`.
So inside `t` we can safely detect what translation framework is used and only temporary patch `$localize` while evaluating it.

# Option 2: Drop using the injection context within `$localize`

We currently use the injection context inside, to save key and value in the `si-translate-service` so that we can use it later to find overrides or falling back to the default value. Based on the configuration `$localize` either returns the key, the key override if available, or the default value.

We could change the behavior of `$localize` to be a pure function. In that case we could no longer return a string, but an object containing the parsed details, like:
```json
{"key": "key", "value": "value"}
```

This means we would need to adjust the type `TranslatableString` to also accept that kind of object.

# Evaluation

`Option 2` has one clear benefit: It does not modify our entire code base. But this is the only good thing.

`Option 1` would solve many issues in a reliable way:
[...]

PR #440, opened 2025-07-31
feat(element-translation-ng): add `t`-function to locally override $localize

spike-rabbit:
The `t` function is the first part of solving our translation issues. Since we can directly import it and not consume it from the global scope, we can reliably use it to inject service.
The `t` function locally patches the `$localize`
so it is independent of the current global scope

See #436

This is part 1 of 3:
1. introduce `t` function
2. use `t` function everywhere
3. drop the old override of `$localize`

---


projects/element-translate-ng/translate/t.ts
fh1ch, line 1: While I do like function name of `t` as I agree with @kfenner that things should remain short and we franklöy follow established best practises like GNU gettext's or lodash's `_`, I frankly don't see a reason why the filename also has to be as short, but up to you.

fh1ch (commented):
@spike-rabbit nice work here, thanks a bunch 🙇 

No findings regarding the implementation except for one question. However, besides this:

- There seems to be *some* disagreement regarding the chosen approach, see https://github.com/siemens/element/issues/436#issuecomment-3145338765.
- There seems to be *some* disagreement regarding the function name. As stated, I see it like @kfenner and we should use a short name, hence why we declined `translate` as name, auto-imports only don't work in *certain* IDEs and are also not used that often, but I could frankly also live perfectly with something like `_t` as thsi would allow auto-imports and be even more in-line with e.g. gettext.

However, I'd recommend to discuss both topics offline, since it will be a larger discussion. WDYT?

/cc @timowolf

spike-rabbit:
All discussions sovled in #436 . We also decided offline to stay with `t`. As this name is internal we can also change it later.

PR #441, opened 2025-07-31
refactor(tabs-next): drop `SiTabsetNextComponent.deselect` event

spike-rabbit:
Using this event in application is pointless.
1. It only works with normal tabs. (but one can also subscribe when using routed tabs)
2. There is a much better option available by listening to the tabs directly.
3. The name is really poor, the main event is an active tab change.
4. It only notifies by user interaction, so one cannot use this to generally track the status.
5. preventing a tab activation violates a11y, is not a typical tab behavior and just in general a very bad approach

Since this is a next component, we can just drop it without a breaking change note.

---

### DIFF
--- a/projects/element-ng/tabs-next/si-tab-next.component.ts
+++ b/projects/element-ng/tabs-next/si-tab-next.component.ts
@@ -31,24 +31,7 @@ export class SiTabNextComponent extends SiTabNextBaseDirective implements OnDest
   override readonly active = model(false);
 
   protected selectTabByUser(): void {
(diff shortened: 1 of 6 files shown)


fh1ch (approved):
@spike-rabbit I frankly can't comment on the use-cases for said event but it indeed looks to be quite poorly implemented and the examples are frankly a bit fringe. I therefore have no issues dropping it, especially since this is a `-next` component. Bringing it back based on project feedback would be easy enough.

In terms fo changes, flawless work here, thanks 🙇 

LGTM 👍

PR #442, opened 2025-07-31
e2e: update snapshots and lower threshold

spike-rabbit:
Please read the commits.

---


fh1ch (approved):
@spike-rabbit nice one, thanks a lot 🙇 

We have to see how stable the new value is, but we indeed missed a lot of color updates with the old threshold, so all fair.

LGTM 👍

PR #444, opened 2025-07-31
docs(form): do not apply `col-*` classes directly on `si-form-item`

spliffone:

---

### DIFF
--- a/src/app/examples/datatable/datatable-selection.html
+++ b/src/app/examples/datatable/datatable-selection.html
@@ -47,14 +47,16 @@
       <si-card class="mt-4" heading="Selection type">
         <div class="card-body" body>
           <div class="row">
-            <si-form-item class="col-6" label="Selection type">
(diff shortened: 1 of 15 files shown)


spike-rabbit:
I don't like those changes. Having col-classes on the si-form-item used to work since a long time.

spliffone:
> I don't like those changes. Having col-classes on the si-form-item used to work since a long time.

@spike-rabbit  Let me explain why do we need an extra element: 

- we have an css selector on `.row > *` which introduce a inline-padding to all direct children of the `.row`
- but the `si-form-item` apply the `.form-check` class directly to the host element (which is also correct) and as discussed - an inline-padding-start on the form-check would fix all the visualisation issues we have right now with checkboxes.

I don't see an alternative to resolve it except we adjust specifically for the combination of `.form-check` and `.row > *`

spike-rabbit:
Ah sh**. I remember. You are right.

spike-rabbit (approved):
:+1:

PR #446, opened 2025-07-31
fix(form): ensure default width of si-form-fieldset labels is 16% in …

spliffone:
…horizontal layout

See si-form example `Actions` fieldset

---


spliffone:
@spike-rabbit  The fieldset label in si-form is currently to small (see below) since no variable --si-form-label-width is defined. Basically, it falls back to 16% inline-size which is taken from the `<div class="pt-1 pb-0">` instead of the entire host size.

image

PR #458, opened 2025-08-04
fix(search-bar): do not emit `searchChange` during initialisation

prasepic:

---

### DIFF
--- a/projects/element-ng/search-bar/si-search-bar.component.ts
+++ b/projects/element-ng/search-bar/si-search-bar.component.ts
@@ -117,7 +117,7 @@ export class SiSearchBarComponent implements OnInit, OnDestroy, ControlValueAcce
 
   /** @internal */
   writeValue(value: string): void {
-    this.setSearch(value);
+    this.writeSearchValue(value ?? '');
   }
 
   /** @internal */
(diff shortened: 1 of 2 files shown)


chintankavathia (approved):
👍

projects/element-ng/search-bar/si-search-bar.component.ts
```diff
@@ -151,14 +153,15 @@ export class SiSearchBarComponent implements OnInit, OnDestroy, ControlValueAcce
...
-      // FIXME: in v48, move outside the `if` so we don't emit when `value === undefined`
-      const searchVal = value ?? '';
-      this.searchValue.set(searchVal);
-      this.inputRef().nativeElement.value = searchVal;
-      this.onChange(searchVal);
-      this.searchChange.emit(searchVal);
+      this.searchValue.set(value);
+      this.inputRef().nativeElement.value = value;
+      this.onChange(value);
+      if (!firstChange) {
```
spike-rabbit, line 162: After having a look here, I guess the whole structure of the component needs an overhaul.

form-inputs (like this component) should have a clear flow of events. We have events triggered by user (typically called change) and we have events from the application (typically called write).

- Changes by the user should update every state and cause emitting of the respective events.
- Changes by the application should only effect the internal state and wright to the UI. Never fire any events.

If an app uses NgForm and the inputs in parallel it is their responsibility to update both states in case they write
as we do not emit any change.

`setSearch` is called in both cases which is bad. Instead, please split in two functions:
- searchChange called when user updates something
- writeSearch for application writes (called from ngOnChanges and writeValue)

spike-rabbit (changes requested):

spike-rabbit (approved):

PR #470, opened 2025-08-05
refactor(typescript): use stricter path mappings

spike-rabbit:
This ensures that even during development,
only valid paths are accepted.
Previously paths like this worked as well:
`@siemens/element-ng/entrypoint/file-in-entrypoint.ts`.

Since we have the convention that every entry-point has an `index.ts` we can enforce using it.

---


fh1ch (approved):
LGTM 👍

PR #476, opened 2025-08-05
chore(demo): add missing examples

chintankavathia:


chintankavathia:
> I think I never moved those examples, as they are never used

why do we then even support showing modal with custom angular component 🤷

kfenner:
@chintankavathia @spike-rabbit  Let's get them in 👍

PR #501, opened 2025-08-07
chore(dep): remove @angular/animations from optional dependencies

chintankavathia:
BREAKING CHANGE: `@angular/animations` package is no longer optional. Package was marked optional incorrectly previously but is required for animations to work.


spike-rabbit (approved):
:+1:

PR #504, opened 2025-08-07
fix(dashboards): prevent footer button focus ring from being clipped

chintankavathia:
on flexible dashboard while adding widget focus ring of footer buttons are not fully visible from bottom.

### DIFF
--- a/projects/dashboards-ng/src/components/widget-catalog/si-widget-catalog.component.scss
+++ b/projects/dashboards-ng/src/components/widget-catalog/si-widget-catalog.component.scss
@@ -32,3 +32,11 @@
     margin-inline-start: map.get(variables.$spacers, 6);
   }
 }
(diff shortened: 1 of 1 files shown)


spike-rabbit (changes requested):
I see the problem, but I guess this is not solving the root cause here.
It seems like that there are many useless overflow hiddens / auto that don't have any use.

Dropping those also solves this issue

chintankavathia:
> I see the problem, but I guess this is not solving the root cause here. It seems like that there are many useless overflow hiddens / auto that don't have any use.
> 
> Dropping those also solves this issue

I don't remember exactly but there were some edge cases for which we needed overflow settings. @timowolf May be you know it.

spike-rabbit:
Fixed with #1895

PR #507, opened 2025-08-07
build(release): add next-release workflow

spike-rabbit:
The next-release workfow works slightly different than the regular one. Instead of just releasing from the `ǹext` branch,
we automatically update it to the `main` branch.
In addition we also skip the generation of a changelog.

The reason for those changes are technical restrictions: When trying to reconfigure `semantic-release` for doing next releases on `main`, we face the issue, that `semantic-release` no longer has a main release branch. There seem to be no way to avoid this problem.

Since we cannot release from main, we would need to bring back the release commit to main in order to ensure a linear history.
Yet there is a risk, that someone may merges while we do a release (which takes up to 15min). In that case we could no longer bring back the commits and thus end up in an invalid state. Just skipping the commiting for next releases solves this issue. This is also more aligned with our previous process.

Since we plan to always pre-release that latest `main` the pipeline automatically updates the `next` branch. So we don't need to do this manually, unless there was an important pipeline script change.

The change was successfully tested here: https://github.com/spike-rabbit/element/

I created two pre-release and one major release.
This proves that:
- pre-releases work in general without a commit
- automatic updating the next branch works
- semantic-release creates the changelogs as expected

---


fh1ch (approved):
@spike-rabbit thanks a lot for the work here 🙇 

It's a bit more involved than originally intended, but the changes look good and the workflow makes generally sense, so let's give this one a try.

LGTM 👍

PR #509, opened 2025-08-07
docs(list-details): simplify example

spike-rabbit:
Inline the tabs to increase readability of the example. Since the focus is here on the list-details and not on tabs, we should keep it as simple as possible.

---


fh1ch (approved):
@spike-rabbit makes sense, thanks a lot 🙇 

LGTM 👍

Issue #511, opened 2025-08-07
help-button in si-form-item


spike-rabbit:
### Prerequisites

- [x] I have read the [Contributing Guidelines](https://github.com/siemens/element/blob/main/CONTRIBUTING.md).
- [x] I have not leaked any internal/restricted information like screenshots, videos, code snippets, links etc.

### Feature request

It should be possible to use the help-button in an si-form-item.
The API should look like this:

```html
<si-form-item label="Inpu">
     <button
        si-help-button
        type="button"
        siHelpTitle="Help for input"
        [siHelpContent]="..."
  >
    How to enter proper input values
  </button>
  <input class="form-control" [formControl]="control">
</si-form-item>
```

This should also work with checks.

### Use case

It is a very common pattern to have inputs with additional help texts.

PR #514, opened 2025-08-07
feat(form): support context help button in form-items

spliffone:
See #511

---


spliffone:
> Can you just build ti this way:
> 
> ```
> <div class="form-label">
>   <label>...</label>
>   <button si-help-button>...</button>
> </div>
> ```
> 
> Then there is no need to introduce this form-label-container

That was my first approach but it won't work since the `.required` class only work for `.form-label`. Means Either the star `*` is behind the help button or we need to completely refactor the the label + required styles.

spike-rabbit (approved):
👍

kfenner (approved):
LGTM 👍

PR #516, opened 2025-08-08
build: add start:prod command for serving examples on production build

chintankavathia:
---


spike-rabbit:
Do we have a common use case for this? I tried to keep the package scripts focues on what is actually needed and for prod start I don't see it.

chintankavathia:
> Do we have a common use case for this? I tried to keep the package scripts focues on what is actually needed and for prod start I don't see it.

I cannot generate vrt locally without running demos from dist

spike-rabbit:
But than we also would need the script for the dashboards?

chintankavathia:
> But than we also would need the script for the dashboards?

dashboards already has `dashboards-demo:start:prod`

PR #524, opened 2025-08-08
feat(copyright-notice): use the company-name provided by the theme

spike-rabbit:
A theme typically belongs to a certain company.
So it makes sense to also include the company name there. We anyway already have the logo and logo label in the theme which are typically also company related.

---


projects/element-ng/copyright-notice/si-copyright-notice.component.spec.ts
```diff
@@ -48,7 +48,6 @@ describe('SiCopyrightNoticeComponent', () => {
     const copyrightElement = element.querySelector<HTMLElement>('si-copyright-notice');
     expect(copyrightElement).toBeTruthy();
 
-    expect(copyrightElement?.innerText).toContain('Sample Company');
```
spike-rabbit, line 51: CSS content does not show up in `innerText`, so we can no longer check it here

spike-rabbit (commented):
to be able to test this in the demo, you currently need to set the `--element-company-name` manually as the siemens does not yet has it.

One also need to remove the input in the si-copyright-notice

projects/element-ng/copyright-notice/si-copyright-notice.component.html
```diff
@@ -1 +1,5 @@
-<p>© {{ copyrightInfo() }}</p>
+<p>
+  ©
+  <span class="company">{{ company() }}</span>
```
timowolf, line 3: This is not a good idea, as copyrights follow different formatting. For example, check the reuse tool and you find other best practises. The copyright should be provided as a string. This brings most flexibility.

timowolf (dismissed):

projects/element-ng/copyright-notice/si-copyright-notice.component.spec.ts
kfenner: Please keep this test and add `Sample Company` explicitely here in the provider above. We should still test this functionality in the unit test. The CSS variable replacement can be tested via VRTs, ok?

kfenner:
> to be able to test this in the demo, you currently need to set the `--element-company-name` manually as the siemens does not yet has it.
> 
> One also need to remove the input in the si-copyright-notice

What do you mean with removing the input??? We want to keep the input!!!

spike-rabbit:
> > to be able to test this in the demo, you currently need to set the `--element-company-name` manually as the siemens does not yet has it.
> > One also need to remove the input in the si-copyright-notice
> 
> What do you mean with removing the input??? We want to keep the input!!!

This was not exactly clear in my comment. I am only refering to the example of the `si-copyright-detials`. In order to see there the name from the them, you have to drop the input. But in general no input is dropped.

projects/element-ng/copyright-notice/si-copyright-notice.component.spec.ts
spike-rabbit: done

kfenner (approved):
Awesome, this is genius work, thank you 🚀 🎉

PR #542, opened 2025-08-11
build(element-translate-ng): only export `$localize` global definitio…

spike-rabbit:
…n as type

Previously a broken output was generated.
Since no actual code was generated no `.js` is generated when compiling. But the generated `package.json` was referencing one that never exists. With this change, we only export the `d.ts`.
A positive side-effect is, that we can now use the package name in our ts-configs as they can see the export in our src `package.json`, which is linked into the `node_modules` by npm workspaces.

---


fh1ch (approved):
LGTM 👍

PR #550, opened 2025-08-12
fix(list-details): make back button outline complete on focus

Killusions:
Need a new stacking context so the focus outline of the back button also shows when not hovering

---


Killusions:
@spliffone

PR #563, opened 2025-08-13
feat(maps): add maps module

chintankavathia:


projects/maps-ng/src/components/si-map/shared/themes/element.ts
```diff
@@ -0,0 +1,49 @@
...
+      colorPalette: {
+        status: [status.info, status.success, status.warning, status.danger],
+        element: [
+          getProp(style, '--siemens-red-500'), // $siemens-red-500
+          getProp(style, '--siemens-orange-700'), // $siemens-orange-700
+          getProp(style, '--siemens-yellow-300'), // $siemens-yellow-300
+          getProp(style, '--siemens-green-500'), // $siemens-green-500
+          getProp(style, '--siemens-blue-500'), // $siemens-blue-500
+          getProp(style, '--element-petrol'), // $element-petrol
+          getProp(style, '--siemens-deep-blue-500') // $siemens-deep-blue-500
```
spike-rabbit, line 42: Why not using the tokens everywhere? '--siemens-*' are no css vars

projects/maps-ng/src/components/si-map/shared/themes/style.ts
```diff
@@ -0,0 +1,2905 @@
...
+const boldFonts = ['Siemens Sans'];
+
+// Colors
+const siemensDeepBlue900 = 'rgba(32, 33, 50, 1)';
+const siemensGreen700 = (opacity: number): string => `rgba(28, 112, 63, ${opacity})`;
+const siemensGreen500 = (opacity: number): string => `rgba(40, 191, 102, ${opacity})`;
+const siemensGreen300 = (opacity: number): string => `rgba(114, 230, 163, ${opacity})`;
+const siemensBlue700 = 'rgb(30, 82, 153)';
+const siemensBlue100 = 'rgb(210, 226, 247)';
+
```
spike-rabbit, line 19: I think we need to consumer all of those from the css vars.
Same for every location in this file

src/app/mocks/custom-shape-style.ts
```diff
@@ -0,0 +1,54 @@
+/**
+ * Copyright (c) Siemens 2016 - 2025
+ * SPDX-License-Identifier: MIT
+ */
+export const shapes: any = {
+  systems: {
+    fillcolor: '#b55cb8',
+    strokecolor: '#bf79c2',
```
spike-rabbit, line 8: Did you check those colors? Can we also use getProp here?

spike-rabbit (changes requested):
Do we have any other options with maptiler key / url.

Would be nice if we can just keep using --define.
Maybe just use an env var in the package.json start script

projects/maps-ng/src/components/si-map/shared/themes/style.ts
chintankavathia: I am not sure how to handle this as some of the rgb are not part of our tokens. e.g `siemensDeepBlue900` here is #202132 which is no where in our tokens. 🤷

esbuild-env-plugin.mjs
```diff
@@ -0,0 +1,73 @@
+import fs from 'fs';
```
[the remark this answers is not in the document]
chintankavathia: same as https://github.com/siemens/element/blob/main/projects/live-preview/esbuild-component-loader.mjs

spike-rabbit (approved):

PR #568, opened 2025-08-13
refactor(filtered-search): add default text for item count

spike-rabbit:
This is a clean-up of the deprecation removal.
Before the deprecation removal, the `items` input
was used to generate the text for multi-selects by default. After the removal, we forgot to add a default text to for the `itemCountText`.

---


spliffone:
We should drop the `itemCountText` input for the test wrapper since we use the same value. See:
https://github.com/siemens/element/blob/e23a5bd43a23f68839d3ac583e179b826bc5ada8/projects/element-ng/filtered-search/si-filtered-search.component.spec.ts#L55

spike-rabbit:
> We should drop the `itemCountText` input for the test wrapper since we use the same value. See:
> 
> https://github.com/siemens/element/blob/e23a5bd43a23f68839d3ac583e179b826bc5ada8/projects/element-ng/filtered-search/si-filtered-search.component.spec.ts#L55

This is not possible. We mocked the translate service, so without overriding  the itemCountText just the key is shown without the count.

spliffone (approved):

kfenner (approved):
LGTM 👍

PR #571, opened 2025-08-14
feat(list-details): support router integration

spike-rabbit:

---


Killusions:
I found one unrelated bug (also happens without the router) that happens when a child is selected while the window resizes to mobile, after, the table is too small, maybe we need to send a listWidthChange later so the table can resize, because it looks like the resize-observer does not trigger.

ScreenFloat Shot of Arc on 2025-08-15 at 01-29-25

Killusions:
And I forgot to say but I also like the code and API 👍

Killusions (approved):

kfenner (approved):
LGTM 👍

PR #573, opened 2025-08-14
build: add postversion script to charts, dashboards, livepreview and nativecharts

chintankavathia:


spike-rabbit:
As written in map-ng PR, I was wrong. We don't need this.

PR #576, opened 2025-08-14
docs(datatable): modify example for translating headers

prasepic:

---

### DIFF
--- a/src/app/examples/datatable/datatable-row-dragging.html
+++ b/src/app/examples/datatable/datatable-row-dragging.html
@@ -16,35 +16,56 @@
       </ng-template>
     </ngx-datatable-column>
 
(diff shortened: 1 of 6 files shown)


spike-rabbit (changes requested):
I also think, that there is no need to show translation with the column input.
One should basically always use the declarative approach with translation.
Otherwise the code is so much harder to read.
Even for server loaded columns it would be nicer, to just use a @for loop in the template

spike-rabbit:
Ah sorry, is was not very specific, I was also having this in my mind as @chintankavathia :

` <ngx-datatable-column name="Name | tanslate" />`.

I guess it would make sense to just update an exisitng example.

Please use `docs(datatable):...` as a commit message

spike-rabbit (approved):
👍

PR #591, opened 2025-08-19
fix(list-details): hide empty header

Killusions:
Currently the height is fixed (inherited from nav-tab) and the border is always shown, but if there is no content and no title and we're in desktop mode (not mobile), we should completely hide it.

---


spike-rabbit:
@panch1739 / @hbxes does the design allow, the have a detail section without a header?

panch1739:
> To be more precise. Currently the details pane of the list-details always has a header.
> 
> @panch1739 / @hbxes does the design allow, the have a detail pane without a header?

@spike-rabbit @Killusions sorry for the late reply! I was on holidays when this PR was created and i never saw the question. Yes, we should totally allow the list-detail also without header. Yes, you get into questions like how to put the node name in it, but we need to provide teams with flexibility.

PR #592, opened 2025-08-19
fix(list-details): prevent overflow

Killusions:
Currently, if a user fails to handle overflow themselves in either the list or details, it creates a buggy scrolling behavior, which is confusing for the user to figure out and solve. Adding overflow-y: auto to the two body components should not have any side effects (still works for flex layout or if there's a nested scrollable) but prevent this weird behavior.

---


spike-rabbit:
We really should make the application deal with the scrolling. 
We never really know what they need. Bootstrap is also not just adding overflows to card bodys and alike. Neither should we.

PR #602, opened 2025-08-19
refactor(list-details): use flex-basis

spike-rabbit:
Previously, the `listSize` supported a value of `default` beside a percentage valued, which was causing issues, as the `listSize` was not reliably a number.
With this change, this `listSize` is now always a number.

---


fh1ch (approved):
@spike-rabbit make sense, thanks a bunch, the demo also works just fine locally, so all good 🙇 

LGTM 👍

PR #606, opened 2025-08-20
refactor(form): use same spacing/line-height as in figma

spike-rabbit:
This is a preparation for the help pattern where we need the line-height of 16px (default).

---


kfenner (approved):
LGTM 👍

PR #609, opened 2025-08-20
refactor(form): use `col-form-label` in inline mode

spike-rabbit:
Previously we overrode the `.form-label` when
the form-item was switched to the inline mode.
With this change, we just use the bootstrap standard class.

---

---

[Documentation](https://d33c9dcnqinn2a.cloudfront.net/pr-609/pages/).
[Examples](https://d33c9dcnqinn2a.cloudfront.net/pr-609/pages/element-examples/).
[Dashboards Demo](https://d33c9dcnqinn2a.cloudfront.net/pr-609/pages/dashboards-demo/).
[Playwright report](https://d33c9dcnqinn2a.cloudfront.net/pr-609/playwright-report/).

**Coverage Reports:**

Code Coverage

- [element-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-609/pages/coverage/element-ng/)
- [element-translate-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-609/pages/coverage/element-translate-ng/)

### DIFF
--- a/projects/element-ng/form/form-fieldset/si-form-fieldset.component.html
+++ b/projects/element-ng/form/form-fieldset/si-form-fieldset.component.html
@@ -1,4 +1,9 @@
-<div class="form-label d-flex gap-2">
+<div
+  class="d-flex gap-2"
+  [class.form-label]="!container?.isInline()"
+  [class.col-form-label]="container?.isInline()"
+  [class.py-1]="container?.isInline()"
+>
   <span [class.required]="isRequired()" [id]="labelId">{{ label() | translate }}</span>
   <ng-content select="[si-help-button]" />
 </div>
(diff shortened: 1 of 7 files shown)


dr-itz:
> The CSS Angular generates for all the `:host-context` is absolutely horrible. So that needs addressing. However we can simply invert the logic here, i.e. only treat the `xs` case as special. I did that in #1756.
> 
> Much smaller CSS and a pure CSS solution.
> 
> (At some point we might consider container queries instead and drop the resize observer from the forms 🤷 )

so...that breaks because when using si-form-item w/o a si-form-container...

spike-rabbit:
We will go the container-query route with the new form-field. For form-item we probaply need to keept it as is and just fix the CSS.

PR #611, opened 2025-08-20
refactor(form): adjust label padding

spike-rabbit:
This is a preparation for the help button.
It prevents the focus outline from overlapping the input.

---


panch1739:
> @spike-rabbit Thank you for the change, this looks good for a start.
> 
> @panch1739 @hbxes With this PR, we adjust the spacing for the form labels to 4px top and 4px bottom (spacer-2) to make room for the help icon. Dani actually designed it in Figma as 2px top and 6px bottom. We thought we want to go with minimal impact for now as we saw that we also do not follow the designed spacing between the form items (24px in Figma, 12px in implementation). I'll create a separate issue for this so we can go together through this again and make the final adjustments in a follow-up.

@kfenner this is awesomeeee, i had talked about this with @spike-rabbit previously because i really wanted to update the spacing. We even had this discussion when we removed the 8px spacing on the left. I'll update the figma component, i think this is the way to go.

Regarding spacing in the forms, 12px is ok....i think we can recommend 12 or 24, depending on the spacing needs.

PR #613, opened 2025-08-20
fix(forms): always require a label in `si-form-item`

spike-rabbit:
Not having a label is an a11y flaw.
This was not enforced before to maintain compatibility with the legacy mode. Since the legacy mode was officially dropped we no longer need this.

BREAKING CHANGE: The `si-form-item` now always requires a label. Not having a label is an accessibility flaw.

---


spike-rabbit:
merged as part of another PR

PR #634, opened 2025-08-22
Build/update maps version properly on release

spike-rabbit:
…before

Maps currently only specifies a major version.
So when doing a release, we should not override this with a specific version.

This PR also adds maps to the workspace.

---


kfenner (approved):
LGTM 👍

PR #635, opened 2025-08-22
test(maps): use dummy fallbacks for color tokens

chintankavathia:
when tests runs they cannot find our tokens and falls back to empty string which openlayer doesn't like and breaks.


spike-rabbit (approved):
👍

PR #642, opened 2025-08-22
fix(formly): link error messages to custom controls

spike-rabbit:
The automatic linking by the `si-form-item` does not work due to broken dependency injection.
So we must link it ourselves.

In the long run, we should stop using the `si-form-item` in `si-formly`.

---


kfenner (approved):
LGTM 👍

PR #659, opened 2025-08-28
refactor(element-ng): use OnPush change detection in remaining components

chintankavathia:
following components should use `ChangeDetectionStrategy.OnPush`:

- si-badge
- si-login-single-sign-on
- si-password-strength
- si-status-counter

## Summary by CodeRabbit

* **Refactor**
  * Optimized change detection strategy across multiple components for improved performance and efficiency. No visible changes to user interface or functionality.

### DIFF
--- a/projects/element-ng/badge/si-badge.component.ts
+++ b/projects/element-ng/badge/si-badge.component.ts
@@ -2,7 +2,7 @@
  * Copyright (c) Siemens 2016 - 2025
  * SPDX-License-Identifier: MIT
  */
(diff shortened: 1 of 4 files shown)


chintankavathia:
> @chintankavathia When searching, there are many more components that are not using `OnPush`. Shall all of them be updated or what should be the strategy here?

we can only update these ones others would be a breaking change technically so we can plan for `v49`

spike-rabbit (changes requested):
LGTM, but please make this a refactor as this has no consumer impact.

PR #661, opened 2025-08-28
chore: set OnPush as default change detection strategy for components

chintankavathia:
Update angular.json schematics configuration to set OnPush change detection as default for component generation.


kfenner (approved):
Niice 🎯 🙇

PR #662, opened 2025-08-28
docs(help-patterns): add link to help button

panch1739:
This PR adds a link to the help button example in playground. As im not sure how to move the code, at least it would be nice to have a direct link to the example :)

---

### DIFF
--- a/docs/patterns/help.md
+++ b/docs/patterns/help.md
@@ -47,7 +47,8 @@ The tour usually highlights key elements of the interface and provides brief ins
 
 ### Contextual help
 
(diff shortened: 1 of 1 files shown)


kfenner:
@panch1739 I think we should add the "Help button" as a separate chapter (proably under "Buttons & menus") in the docs and add there a code section with the example. It is really too hidden and just adding this link under "Design" is not cutting it for me.

I would call it "help button" as the directive is called `si-help-button` to stay in line. WDYT?

In the forms chapter, it's already described how to add the help button:
https://element.siemens.io/components/forms-inputs/forms/#context-help

PR #666, opened 2025-08-29
chore(demo): rename WebpackTranslateLoader to CustomTranslateLoader

chintankavathia:
name webpack makes no sense as we don't use webpack builder anymore


spike-rabbit (approved):
👍

PR #167, opened 2025-07-01
docs(README): use absolute urls for links

prasepic:
The license links in npmjs.com is invalid since it is relative.
Other libs also have an absolute URL (ngx-datatable, ionic).

> Describe in detail what your merge request does and why. Add relevant
> screenshots and reference related issues via `Closes #XY` or `Related to #XY`.

---

- [x] I confirm that this MR follows the [contribution guidelines](https://github.com/siemens/element/blob/main/CONTRIBUTING.md).

PR #669, opened 2025-08-29
docs: use correct url path for license.md

chintankavathia:


kfenner:
> I think all these need to be absolute URLs (npmjs will point towards an invalid file)
> 
> see #167

@prasepic You're right, good point! 
@chintankavathia Thanks for the fix, can you change to absolute URLs 🙏

kfenner (approved):
LGTM 👍

PR #683, opened 2025-09-02
refactor(tree-view): use siTreeViewItem for custom templates

chintankavathia:
DEPRECATED: `siTreeViewItemTemplate` is deprecated in favor of `siTreeViewItem` and will be removed in a future version.

Migration Guide:

Before (deprecated):
```html
<si-tree-view>
  <ng-template siTreeViewItemTemplate="root" let-item>
    <div class="custom-item">Root {{ item.name }}</div>
  </ng-template>
  <ng-template siTreeViewItemTemplate="child" let-item>
    <div class="custom-item">Child {{ item.name }}</div>
  </ng-template>
</si-tree-view>
```

After (recommended):
```html
<si-tree-view>
  <ng-template siTreeViewItem let-item="treeItem">
    <si-tree-view-item>
      <div class="custom-item">
      @if (item.level === 0) {
        Root {{ item.name }}
      } @else {
        Child {{ item.name }}
      }
      </div>
    </si-tree-view-item>
  </ng-template>
</si-tree-view>
```


kfenner:
@chintankavathia TBH, both approaches have pros and cons. E.g. if you have many many different templates, having a large switch case inside the template is not so great as it will have to be evaluated on each node. But we're doing similar stuff elsewhere, so not sure how big of a performance impact it has.

If this issue is neglectible, then I'd also vote to deprecate the other one.

spike-rabbit:
I fully agree with this one.

Since you anyway wrote it already, please copy the `Overview` part of the issue description also into the commit and I think it makes sense to rephrase the first line of the deprecation note:

`DEPRECATED: siTreeViewItemTemplate is deprecated in favor of siTreeViewItem`.  

The second part felt weird to me when reading. I first understood that the `siTreeViewItem` directive will be removed

kfenner (approved):
As my concerns seem no longer valid, I'm fine with this, let's merge 👍

projects/element-ng/tree-view/si-tree-view-item-template.directive.ts
```diff
@@ -6,6 +6,45 @@ import { Directive, inject, input, TemplateRef } from '@angular/core';
...
+ * ```html
+ * <si-tree-view>
+ *   <ng-template siTreeViewItem let-item="treeItem">
+ *     <si-tree-view-item>
+ *      <div class="custom-item">
+ *       @if (item.level === 0) {
+ *         Root {{ item.name }}
+ *       } @else {
+ *         Child {{ item.name }}
+ *       }
```
spike-rabbit: I guess just using the level in the name is also fine

spike-rabbit (approved):
👍

PR #685, opened 2025-09-02
fix(threshold): ensure number input and buttons are above other elements

mistrykaran91:
Currently, the number input and button is overriden by div which disable half of the increment button to be clicked. This fix will make sure that number input and increment and decrement buttons are above.

---


dr-itz (approved):
👍 makes sense. Caused by the negative margin in horizontal mode. I wonder why this hasn't been noticed before

kfenner (approved):
LGTM 👍

PR #687, opened 2025-09-03
refactor(datatable): use `siDatatableInteraction` only with ngx-datatable

chintankavathia:
allow `siDatatableInteraction` to be applied only on `ngx-datatable` and use direct `DatatableComponent` instance values for `selectionType` and `virtualization`.

---


kfenner (approved):
@chintankavathia Good idea, however, the new version will actually break when we migrate to signals. WDYT, do you want to make it already compatible with signals now, anticipating we want to support a larger supported version range, or would we only handle this after signals migration of datatable?

kfenner (approved):
LGTM 👍

PR #704, opened 2025-09-08
refactor(tree-view): use better input names

chintankavathia:
DEPRECATED: following inputs from `SiTreeViewComponent` are now deprecated with replacement.

- `folderStateStart` (use `folderStateEnd`)
- `isVirtualized` (use `disableVirtualization`)
- `enableStateIndicator` (use `hideStateIndicator`)
- `enableIcon` (use `hideIcon`)
- `enableContextMenuButton` (use `hideContextMenuButton`)
- `inheritChecked` (use `preventInheritChecked`)

Note that replaced inputs have inverted behavior. eg,

// Before
`<si-tree-view  [enableIcon]="false">`

// After
`<si-tree-view  hideIcon>` or `<si-tree-view  [hideIcon]="true">`

---


kfenner:
@chintankavathia Thank you for this. However, I think we shouldn't do this as there is not too much value for the users and then we have lots of deprecations just because of the name (and default value). We should first think about the future of the tree and whether we might want to switch to a new implementation (using cdk, perhaps) for a declarative tree and then making a factory for that which is closer to the API of the current tree. And then finally deprecating the whole current tree alltogether.

I just don't see a priority for this atm. as almost nobody complaints about the current tree and it is very stable and works in huge applications.

/cc @spike-rabbit @timowolf

PR #705, opened 2025-09-08
feat(filter-bar): add disableAutoTranslation option to avoid unwanted translation

mistrykaran91:
When user types a word that matches the translation keys then filter pill displays the actual translated value.
Using `disableAutoTranslation` will prevent automatic translation of filter pill text when it matches translation keys

---

### DIFF
--- a/projects/element-ng/filter-bar/filter.ts
+++ b/projects/element-ng/filter-bar/filter.ts
@@ -25,4 +25,10 @@ export interface Filter {
    * Status of filter pill
    */
   status: FilterStatusType;
+  /**
+   * Disables automatic translation of title and description text.
+   * When set to true, the title and description will be displayed as-is without
+   * attempting translation.
+   */
+  disableAutoTranslation?: boolean;
 }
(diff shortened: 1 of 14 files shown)


akashsonune (changes requested):

projects/element-ng/filter-bar/filter.ts
```diff
@@ -25,4 +25,10 @@ export interface Filter {
    * Status of filter pill
    */
   status: FilterStatusType;
+  /**
+   * Disables automatic translation of title and description text.
+   * When set to true, the title and description will be displayed as-is without
```
kfenner, line 30: Please see in the comments above, we normally use links to refer to other properties when we do proper comments 👍 

Also, please make sure the `title` and `description` fields are marked as `TranslatableString` types instead of plain `string`.

```suggestion
   * Disables automatic translation of {@link title} and {@link description} text.
   * When set to true, the {@link title} and {@link description} will be displayed as-is without
```

kfenner:
@siemens/siemens-element-members Please review the approach here. Do you think this is a good solution to prevent user input from accidentially being translated when the entered text matches a TRANSLATION.KEY?

kfenner (changes requested):

spike-rabbit:
Please see #758. We plan to have a common marker instead of component specific solution.

PR #714, opened 2025-09-11
fix(search-filter-value): multiple clicks required for clearing a sea…

shubham1121:
Fixed two clicks required for clearing out a search filter in si-filtered-search component.

Fixes : [si-filtered-search component Search field search cross button needs to be clicked twice while clearing the filter.](https://code.siemens.com/simpl/simpl-element/-/issues/2559)

### DIFF
--- a/projects/element-ng/filtered-search/si-filtered-search-value.component.ts
+++ b/projects/element-ng/filtered-search/si-filtered-search-value.component.ts
@@ -168,15 +168,9 @@ export class SiFilteredSearchValueComponent implements OnInit {
   }
 
   protected clear(): void {
-    if (!this.active() || !this.value().value?.length) {
-      this.deleteCriterion.emit({ triggerSearch: true });
-      return;
-    }
-
-    this.value.update(v => ({
-      ...v,
-      dateValue: undefined,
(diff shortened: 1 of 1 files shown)


timowolf (commented):
Many thanks for the PR. Unfortunately, you removed the feature that was included:

Having an active option, pressing the clear button should remove the content from the input box and not delete the criterion. After your change, the behavior of a multi-select dropdown is also affected. You’ve identified a bug, but the issue is that the text should be removed from the input box, not the criterion.

shubham1121:
@timowolf 
Okay, I will correct the functionality, and text should be removed on single click of that cross button. and if again that button is clicked then it should remove the criterion.

Am I correct here?

spike-rabbit:
THX a lot for your contribution @shubham1121 this is highly appreciated :100: 

As written by @timowolf, if a criterion has a value and is active (focussed) a click on the clear button should clear the input only and not the whole criteria. This is indeed currently broken for normal / typeahead based  criteria. Since the fix is slightly more complicated I fixed it in #781 already. Please feel free to try the behavior there and provide feedback.
If the new PR makes sense for you, please close this one.

PR #716, opened 2025-09-12
feat(tabs): provide canActivate and canDeactivate guards

chintankavathia:
added `SiTabComponent.canActivate` and `SiTabComponent.canDeactivate` to control activation/deactivation of tabs similar to angular router guards.

### DIFF
--- a/projects/element-ng/tabs/si-tab.component.ts
+++ b/projects/element-ng/tabs/si-tab.component.ts
@@ -2,7 +2,7 @@
  * Copyright (c) Siemens 2016 - 2025
  * SPDX-License-Identifier: MIT
  */
-import { ChangeDetectionStrategy, Component, model, OnDestroy } from '@angular/core';
+import { ChangeDetectionStrategy, Component, input, model, OnDestroy } from '@angular/core';
 import { SiIconComponent } from '@siemens/element-ng/icon';
 import { SiTranslatePipe } from '@siemens/element-translate-ng/translate';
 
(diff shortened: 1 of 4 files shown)


panch1739:
> Many thanks, while the implementation is great, can you please also add the canActivate to the example. E.g. Does it make sense to show a tab that cannot be activated? Or should it be hidden @panch1739

If I understand correctly, when a tab/route is protected by canActivate, the user may click it and nothing happens because they don’t have access. From a UX perspective this feels broken, since the system is silently blocking navigation.

I would suggest: 
- By default, tabs should be hidden when the feature is irrelevant to the user’s role.
- If there’s a clear reason for the user to see the tab (for example: to request access, or contact an admin), the tab may remain visible. In this case, navigation should still work but lead to a screen with an [info page](https://element.siemens.io/components/pages/info-page/) component with something like:
“You don’t have access to this feature.
“Contact your administrator to request access.”

@timowolf @chintankavathia would that be possible? maybe we can add a short guide?

chintankavathia:
> > Many thanks, while the implementation is great, can you please also add the canActivate to the example. E.g. Does it make sense to show a tab that cannot be activated? Or should it be hidden @panch1739
> 
> If I understand correctly, when a tab/route is protected by canActivate, the user may click it and nothing happens because they don’t have access. From a UX perspective this feels broken, since the system is silently blocking navigation.
> 
> I would suggest:
> 
> * By default, tabs should be hidden when the feature is irrelevant to the user’s role.
> * If there’s a clear reason for the user to see the tab (for example: to request access, or contact an admin), the tab may remain visible. In this case, navigation should still work but lead to a screen with an [info page](https://element.siemens.io/components/pages/info-page/) component with something like:
>   “You don’t have access to this feature.
>   “Contact your administrator to request access.”
> 
> @timowolf @chintankavathia would that be possible? maybe we can add a short guide?

@panch1739 sorry for the confusion. Apps can provide the feedback (e.g navigate to info page or show a toast message etc) when tab is clicked but is protected by canActivate.

chintankavathia:
> > > > Many thanks, while the implementation is great, can you please also add the canActivate to the example. E.g. Does it make sense to show a tab that cannot be activated? Or should it be hidden @panch1739
> > > 
> > > 
> > > If I understand correctly, when a tab/route is protected by canActivate, the user may click it and nothing happens because they don’t have access. From a UX perspective this feels broken, since the system is silently blocking navigation.
> > > I would suggest:
> > > 
> > > * By default, tabs should be hidden when the feature is irrelevant to the user’s role.
> > > * If there’s a clear reason for the user to see the tab (for example: to request access, or contact an admin), the tab may remain visible. In this case, navigation should still work but lead to a screen with an [info page](https://element.siemens.io/components/pages/info-page/) component with something like:
> > >   “You don’t have access to this feature.
> > >   “Contact your administrator to request access.”
> > > 
> > > @timowolf @chintankavathia would that be possible? maybe we can add a short guide?
> > 
> > 
> > @panch1739 sorry for the confusion. Apps can provide the feedback (e.g navigate to info page or show a toast message etc) when tab is clicked but is protected by canActivate.
> 
> Ann thanks @chintankavathia! Would it be possible to add this sort of guideline somewhere? I think it would be nice to mention that...
> 
> * By default, tabs should be hidden when the feature is irrelevant to the user’s role.
> * If there’s a clear reason for the user to see the tab (for example: to request access, or contact an admin), the tab may remain visible. In this case, navigation should still work but lead to a screen with an [info page](https://element.siemens.io/components/pages/info-page/) component with something like:
>   “You don’t have access to this feature.
>   “Contact your administrator to request access.”
> 
> @timowolf what do you think?

@panch1739 @timowolf added some explanation here https://github.com/siemens/element/pull/735/files let me know if this works

PR #717, opened 2025-09-12
chore(formly): set OnPush as default for formly wrappers

prasepic:

---


timowolf (commented):
Many thanks for the contribution. Why did you not change the `SiFormlyWrapperComponent` component?

I also do not really understand why you did this change. We also have many other components in the folders fields and structural. Why not changing all or does this not make sense?

PR #742, opened 2025-09-18
fix(filtered-search): filter pill overlay stays open opn focus out scenario

### LINKED ISSUE #741: Filtered-search: pill overlay stays open and misaligned
### Prerequisites

### Affected component

_No response_

### Summary

Image

### Version

_No response_

### Steps to reproduce

1. Go to multiselect example - https://element.siemens.io/element-examples/#/overview/si-filtered-search/si-filtered-search-multiselect?q=filtered
2. Delete free text 'Max' and add new criteria 'Location'
3. Select any option from multi options
4. While the overlay is open, click on clear button of previous criteria - Country
5. Note that the overlay pill pf multi select is still open and also not aligned to its criteria

### Code reproduction example

https://element.siemens.io/element-examples/#/overview/si-filtered-search/si-filtered-search-multiselect?q=filtered

Pill should close of focus out

Pill does not close on focus out

akashsonune:
Fixes #741

### DIFF
--- a/projects/element-ng/filtered-search/si-filtered-search-value.component.ts
+++ b/projects/element-ng/filtered-search/si-filtered-search-value.component.ts
@@ -105,6 +105,12 @@ export class SiFilteredSearchValueComponent implements OnInit {
     }
   }
 
+  closeOverlay(): void {
+    if (this.active()) {
+      this.active.set(false);
+    }
+  }
+
   edit(field?: 'value' | 'operator'): void {
     if (this.readonly()) {
       return;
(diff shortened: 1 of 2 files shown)


chintankavathia:
This is not only related to this particular usecase but the way our overlay's currently behave on resizing. e.g if the dropdown is open and if you scroll on viewport the overlay will not reposition itself to element to which it is connected to.

this happens because by default cdk overlay doesn't have any scroll strategy (https://material.angular.dev/cdk/overlay/api#ScrollStrategyOptions) . We should decide which scroll strategy we want to use as this problem is applicable to other components as well (e.g datepicker, typeahead, si-select etc)

timowolf (approved):
Many thanks, I think this helps.

Issue #668, opened 2025-08-29
Use OnPush change detection strategy on remaining components


chintankavathia:
# OnPush Change Detection Strategy Implementation

## Overview
We marked most of our element components with `OnPush` change detection strategy over a year ago. However, several components are still missing this optimization and need evaluation for `OnPush` implementation.

## Components Requiring OnPush

- `si-header-navigation`
- `si-header-selection-item`
- `si-breadcrumb`
- `si-breadcrumb-router`
- `si-card`
- `si-footer`
- `si-help-button`
- `si-inline-notification`
- `si-tooltip`
- `si-content-action-bar`
- `si-side-panel`
- `si-side-panel-content`
- `si-dashboard`
- `si-dashboard-card`
- `si-link-widget`
- `si-list-widget`
- `si-list-widget-body`
- `si-list-widget-item`
- `si-timeline-widget`
- `si-timeline-widget-body`
- `si-timeline-widget-item`
- `si-value-widget`
- `si-value-widget-body`
- `si-widget-base`
- `si-calendar-body`
- `si-initial-focus`
- `si-date-range`
- `si-datepicker`
- `si-datepicker-overlay`
- `si-file-uploader`
- `si-image-cropper-style`
- `si-filter-bar`
- `si-filter-pill`
- `si-filtered-search`
- `si-sort-bar`
- `si-form-container`
- `si-form-validation-tooltip`
- `si-header-dropdown-items-factory`
- `si-icon-status`
- `si-input-pill`
- `si-password-toggle`
- `si-language-switcher`
- `si-menu-factory`
- `si-navbar-vertical`
- `si-navbar-vertical-divider`
- `si-navbar-vertical-group`
- `si-navbar-vertical-header`
- `si-navbar-vertical-item-legacy`
- `si-navbar-item`
- `si-navbar-primary`
- `si-popover` (both legacy and new versions)
- `si-electron-titlebar`
- `si-result-details-list`
- `si-readonly-threshold-option`
- `si-threshold`
- `si-toast-notification`
- `si-tree-view`
- `si-tree-view-item`
- `si-split-part`
- `si-status-bar-item`
#### Formly Components
- `si-formly`
- `si-formly-accordion`
- `si-formly-array`
- `si-formly-object-grid`
- `si-formly-object-plain`
- `si-formly-object`
- `si-formly-object-tabset`
- `si-formly-fieldset`
- `si-formly-horizontal-wrapper`
- `si-formly-icon-wrapper`
[...]

PR #745, opened 2025-09-19
chore(formly): set OnPush as default for structural components

prasepic:
related to #668

---

### DIFF
--- a/projects/element-ng/formly/structural/si-formly-object-grid/si-formly-object-grid.component.ts
+++ b/projects/element-ng/formly/structural/si-formly-object-grid/si-formly-object-grid.component.ts
@@ -2,7 +2,7 @@
  * Copyright (c) Siemens 2016 - 2025
  * SPDX-License-Identifier: MIT
  */
-import { Component, OnInit } from '@angular/core';
+import { ChangeDetectionStrategy, Component, OnInit } from '@angular/core';
 import { FieldType, FormlyModule } from '@ngx-formly/core';
 import { SiFormContainerComponent } from '@siemens/element-ng/form';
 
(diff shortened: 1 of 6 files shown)


projects/element-ng/formly/structural/si-formly-object-grid/si-formly-object-grid.component.ts
```diff
@@ -11,7 +11,8 @@ import { GridColumnConfig, GridRow, ToGridRowConfig } from './si-formly-object-g
 @Component({
   selector: 'si-formly-object-grid',
   imports: [FormlyModule, SiFormContainerComponent],
-  templateUrl: './si-formly-object-grid.component.html'
+  templateUrl: './si-formly-object-grid.component.html',
+  changeDetection: ChangeDetectionStrategy.OnPush
```
timowolf, line 15: Don't we need to add mark for check after `this.rows.push({ classes: rowClasses, columns });`

timowolf (commented):
Many thanks @prasepic, Just one point I am not sure about at.

projects/element-ng/formly/structural/si-formly-object-grid/si-formly-object-grid.component.ts
prasepic: I think that is fine @timowolf.
I have also tested it with our examples.

But let me first verify why it works?

prasepic: works because the method that mutates `rows` is inside `onInit`. Change detection runs anyways after ngOnInit, so we are good

PR #763, opened 2025-09-23
docs(datepicker): improve si-timepicker example

spliffone:

---


akashsonune:
@spliffone Looks cool, Does it make sense to also include here the timepicker-limit [example](https://element.siemens.io/element-examples/#/overview/si-datepicker/si-timepicker-limits?q=time) as this is now a nice playground

akashsonune (commented):
Can we remove the timepicker-limits example now as it is redundant

akashsonune (commented):
Also I noticed one thing, in current example when i change , min time  to 09:30:00, the timepicker will not show validation error. Fine.

Now when I again change the min time back to original 11:30:00, the error should return, but it does not

Image

akashsonune (approved):
LGTM

PR #774, opened 2025-09-25
fix(tabs): always scroll active tab into view

### LINKED ISSUE #772: tabs: an active si-tab should always be visible tablist
### Prerequisites

### Affected component

<si-tab>

### Summary

An active si-tab is not automatically visible in the tablist when the consumer app set the input `active` to true on a tab which at the end of the list. Neither the drop-down show mark the active tab.
 

### Version

@siemens/element-ng@48.0.0

### Steps to reproduce

- create a list of si-tabs with 8 entries and show the label text
- set the last si-tab to active
- shrink the viewport size to a size where not all tabs are visible at once

### Code reproduction example

https://element.siemens.io/element-examples/#/overview/si-tabs/si-tabs

The consumer app can't use the active input on si-tab to make a tab visible in the current list. At the moment this only happens when the user select an tab.

When the consumer app use the active input on si-tab we should ensure that the tab is visible in the current list.

chintankavathia:
Closes #772

### DIFF
--- a/projects/element-ng/tabs/si-tab-base.directive.ts
+++ b/projects/element-ng/tabs/si-tab-base.directive.ts
@@ -82,6 +82,7 @@ export abstract class SiTabBaseDirective implements OnDestroy, FocusableOption {
   /** Event emitter to notify when a tab is closed. */
   readonly closeTriggered = output();
 
+  /** @internal */
   protected readonly tabButton = inject<ElementRef<HTMLElement>>(ElementRef);
   /** @internal */
   readonly tabContent = viewChild('tabContent', { read: TemplateRef });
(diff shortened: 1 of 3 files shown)


spliffone (approved):
I love it and it works!

projects/element-ng/tabs/si-tabset.component.ts
```diff
@@ -68,6 +69,21 @@ export class SiTabsetComponent {
     return tab instanceof SiTabLinkComponent;
   }
 
+  constructor() {
+    effect(() => {
+      if (this.activeTab()?.active() && this.showMenuButton()) {
```
spike-rabbit, line 74:
```suggestion
      if (this.showMenuButton() && this.activeTab()) {
```

I think there is no need to check if the activeTab is active.
Switiching the order should ensure that the effect is triggered less often

spike-rabbit (changes requested):
Please enhance the commit message. `on load` is not really the correct description.

projects/element-ng/tabs/si-tab-base.directive.ts
spike-rabbit, line 136: I think you can make this a public method, and then the element itself can stay protected

spike-rabbit (changes requested):
Found some more little thing

spike-rabbit (approved):
👍

PR #778, opened 2025-09-29
fix(tooltip): dismissible tooltip visibility on Safari

dauriamarco:
Fix for the dismissible tooltip visibility on Safari. Clicking a button **does not give it focus in Safari** according to [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#clicking_and_focus) and [WebKit Bug #22261](https://bugs.webkit.org/show_bug.cgi?id=22261#c68), preventing any focus-dependent logic (e.g., tooltips) from working properly.

---


spike-rabbit:
I am not having a Mac. @dr-itz please confirm here

dr-itz:
> I am not having a Mac.

that's a problem that can be fixed 🤷 😃 

> @dr-itz please confirm here

well, it works as intended. But. On Safari clicking a button doesn't move the focus. That's the platform behaviour. And this MR changes that for buttons with tooltips. I don't like that.

And I also question why would a button show a tooltip on click? Focus by tab is another thing. But click means whatever action is behind the button has already been executed. (except the case like in the demo where the only action is showing the tooltip, but then why the focus trigger).

dauriamarco:
> well, it works as intended. But. On Safari clicking a button doesn't move the focus. That's the platform behaviour. And this MR changes that for buttons with tooltips. I don't like that.
> 
> And I also question why would a button show a tooltip on click? Focus by tab is another thing. But click means whatever action is behind the button has already been executed. (except the case like in the demo where the only action is showing the tooltip, but then why the focus trigger).

@dr-itz @spike-rabbit We had already discussed this fix on GitLab, and I initially also thought about changing the input to `readonly triggers = input<'' | 'click'>();` (or even removing the possibility of having a dismissible tooltip as you mentioned, I see your point). But you correctly pointed out back then that this would be a breaking change.

That’s why I also suggested a non-breaking temporary solution which I don't really like as well, but I haven’t received updates after that. At this stage, we can directly decide whether we want to keep or completely get rid of the dismissible tooltip, and how to handle it (?).

Let me know what you think 👍

spike-rabbit:
I see, first thought that this is still a bug after we updated the keyboard interaction, but this was the popover I thought about.

Anyway, the triggers input is something that should be deprecated and removed. All tooltips are supposed to have an identical interaction pattern: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/tooltip_role. We the description there I see also no Safari problem, correct?

PR #782, opened 2025-09-30
feat(migration): add action dialog migration rule and mappings for v48 update

mistrykaran91:

### DIFF
--- a/projects/schematics/ng-add/action-modal-migration.spec.ts
+++ b/projects/schematics/ng-add/action-modal-migration.spec.ts
@@ -1,0 +1,254 @@
+/**
+ * Copyright (c) Siemens 2016 - 2025
+ * SPDX-License-Identifier: MIT
+ */
+
+import { Tree } from '@angular-devkit/schematics';
+import { SchematicTestRunner } from '@angular-devkit/schematics/testing';
+import * as path from 'path';
+import { beforeEach, describe, expect, test } from 'vitest';
+
+import { addTestFiles, createTestApp, readLines } from '../utils/testing';
+
+const collectionPath = path.join(
+  process.cwd(),
+  'dist/@siemens/element-ng/schematics/collection.json'
+);
+
+describe('action modal migration', () => {
+  let runner: SchematicTestRunner;
+  let appTree: Tree;
+  const name = 'ng-add';
+
+  beforeEach(async () => {
+    runner = new SchematicTestRunner(name, collectionPath);
+    appTree = await createTestApp(runner, { style: 'scss' });
+  });
+
+  const checkTemplateMigration = async (
+    original: string[],
+    expected: string[],
+    sourceFile = `/projects/app/src/components/test/test.component.ts`
+  ): Promise<void> => {
+    addTestFiles(appTree, {
+      [sourceFile]: original.join('\n'),
+      '/package.json': `{
+         "dependencies": {
+          "@simpl/element-ng": "47.0.3",
+          "@simpl/maps-ng": "47.0.3",
+          "@simpl/dashboards-ng": "47.0.3",
+          "@simpl/element-translate-ng": "47.0.3",
+          "some-other-dep": "1.2.3"
+        }
+        }`
+    });
+
+    const tree = await runner.runSchematic('ng-add', { path: 'projects/app/src' }, appTree);
+    const actual = readLines(tree, sourceFile);
+    expect(actual).toEqual(expected);
+  };
+
+  test('should migrate `showAlertDialog` method and its type usage', async () => {
+    const original = [
+      `import { Component, inject } from '@angular/core';`,
+      `import { AlertDialogResult, SiActionDialogService } from '@simpl/element-ng/action-modal';`,
+      `@Component({ selector: 'app-test-action-modal' })`,
+      `export class TestActionModalComponent {`,
+      `showDialog() { inject(SiActionDialogService).showAlertDialog('Message', 'Heading', 'Confirmation button text').subscribe(`,
+      `(value: AlertDialogResult) => { if (value === AlertDialogResult.Confirm) { return 'yes'; } return 'no'; });`,
+      `} }`
+    ];
+
+    const expected = [
+      `import { Component, inject } from '@angular/core';`,
+      `import { AlertDialogResult, SiActionDialogService } from '@siemens/element-ng/action-modal';`,
+      `@Component({ selector: 'app-test-action-modal' })`,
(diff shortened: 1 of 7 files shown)


projects/schematics/siemens-migration/index.spec.ts
```diff
@@ -377,16 +377,16 @@ describe('siemensMigration', () => {
       const dependencies = packageJson?.dependencies;
 
       expect(dependencies?.['@simpl/element-ng']).not.toBeDefined();
-      expect(dependencies?.['@siemens/element-ng']).toBe('48.0.3');
+      expect(dependencies?.['@siemens/element-ng']).not.toBe('47.0.3');
```
spliffone, line 380: Why do we change the version of the siemens migration, this shouldn't be part of the MR

projects/schematics/ng-update/index.spec.ts
```diff
@@ -18,13 +20,190 @@ describe('ng-update migration version 48.0.0', () => {
...
+    const tree = await runner.runSchematic('migration-v48', { path: 'projects/app/src' }, appTree);
+    const actual = readLines(tree, sourceFile);
+    expect(actual).toEqual(expected);
+  };
+
+  describe('action modal migration', () => {
+    test('should migrate `showAlertDialog` method and its type usage', async () => {
+      const original = [
+        `import { Component, OnInit, inject } from '@angular/core';`,
+        `import { AlertDialogResult, SiActionDialogService } from '@simpl/element-ng/action-modal';`,
```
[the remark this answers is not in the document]
mistrykaran91: Should work with that too, I'll add test case

projects/schematics/ng-add/action-modal-migration.spec.ts
```diff
@@ -0,0 +1,266 @@
...
+    });
+
+    const tree = await runner.runSchematic('ng-add', { path: 'projects/app/src' }, appTree);
+    const actual = readLines(tree, sourceFile);
+    expect(actual).toEqual(expected);
+  };
+
+  test('should migrate `showAlertDialog` method and its type usage', async () => {
+    const original = [
+      `import { Component, OnInit, inject } from '@angular/core';`,
```
spliffone, line 53: Why do you need the `OnInit`? I guess we can drop the import

projects/schematics/ng-add/action-modal-migration.spec.ts
```diff
@@ -0,0 +1,266 @@
...
+
+    await checkTemplateMigration(original, expected);
+  });
+
+  test('should migrate `showConfirmationDialog` method and its type usage', async () => {
+    const original = [
+      `import { Component, OnInit, inject } from '@angular/core';`,
+      `import { ConfirmationDialogResult, SiActionDialogService } from '@simpl/element-ng';`,
+      `@Component({ selector: 'app-test-action-modal' })`,
+      `export class TestActionModalComponent {`,
```
spliffone, line 82: Could we simplify the test cases to something like this:
```ts
import { inject } from '@angular/core';
import { ConfirmationDialogResult, SiActionDialogService } from '@simpl/element-ng';`,
const showDialog = (): void => 
inject(SiActionDialogService).showConfirmationDialog('Message', 'Heading', 'Confirmation button text','Decline button text').subscribe(result => console.log(value === ConfirmationDialogResult.Decline))
```

projects/schematics/ng-add/action-modal-migration.ts
```diff
@@ -0,0 +1,167 @@
...
+  const newCode = `${servicePath}.showActionDialog({ ${dialogProperties.join(', ')} })`;
+
+  return {
+    node,
+    newCode,
+    type: 'method-call'
+  };
+};
+
+const createTypeReferenceTransformation = (
```
spliffone, line 107: Can we improve the naming? The current name is really generic but at the end the code is specific  the action dialog.

projects/schematics/ng-add/action-modal-migration.ts
```diff
@@ -0,0 +1,167 @@
...
+  }
+
+  return {
+    node,
+    newCode: `'${matchingReplacement.new}'`,
+    type: 'type-reference'
+  };
+};
+
+const generateDialogProperties = (
```
spliffone, line 127: Can you provide a description / example of the transformation?

projects/schematics/ng-add/action-modal-migration.spec.ts
```diff
@@ -0,0 +1,266 @@
...
+    const tree = await runner.runSchematic('ng-add', { path: 'projects/app/src' }, appTree);
+    const actual = readLines(tree, sourceFile);
+    expect(actual).toEqual(expected);
+  };
+
+  test('should migrate `showAlertDialog` method and its type usage', async () => {
+    const original = [
+      `import { Component, OnInit, inject } from '@angular/core';`,
+      `import { AlertDialogResult, SiActionDialogService } from '@simpl/element-ng/action-modal';`,
+      `@Component({ selector: 'app-test-action-modal' })`,
```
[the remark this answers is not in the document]
mistrykaran91: No its not necessary but I would like to make it as close as real component thats why I have kept only selector.

projects/schematics/ng-add/action-modal-migration.spec.ts
mistrykaran91: I would like to keep the class and component rest I have made the changes

projects/schematics/ng-add/action-modal-migration.spec.ts
mistrykaran91: My bad, I missed to remove it while adding it

projects/schematics/ng-add/action-modal-migration.ts
mistrykaran91: I think this name being inside `action-modal-migration.ts` is quite explanatory that it replaces the types which are imported. However, I have renamed this to `createActionDialogMethodCallTransformation`

projects/schematics/ng-add/action-modal-migration.ts
```diff
@@ -0,0 +1,167 @@
...
+        ACTION_MODAL_SYMBOLS
+      );
+
+      if (!actionModalImports || actionModalImports.length === 0) {
+        continue;
+      }
+
+      const pendingTransformations: CodeTransformation[] = [];
+
+      const visitNodeAndCollectTransformations = (node: ts.Node): void => {
```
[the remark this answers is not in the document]
mistrykaran91: Not sure I got it

PR #785, opened 2025-09-30
chore(uv): fix python version

spike-rabbit:
UV was previously reporting this error:
```
warning: The `requires-python` specifier (`~=3.11`) in `element` uses the tilde specifier (`~=`) without a patch version. This will be interpreted as `>=3.11, <4`. Did you mean `~=3.11.0` to constrain the version as `>=3.11.0, <3.12`? We recommend only using the tilde specifier with a patch version to avoid ambiguity.`
```

So we are now using `>=3.11` as a new selector.

---


fh1ch (approved):
LGTM 👍

PR #791, opened 2025-10-01
refactor(notification-item): improve scroll-area of popover example

ljanner:
The scroll-area of a popover should be the body while the header and footer of the popover are sticky.

---


ljanner:
Keeping as a draft until spacing has been fully clarified.

src/app/examples/si-notification-item/si-notification-item-popover.ts
```diff
@@ -45,6 +45,18 @@ import { LOG_EVENT } from '@siemens/live-preview';
...
+    .notification-dropdown {
+      max-width: none;
+      max-height: 70vh;
+    }
+
+    @media (min-width: 768px) {
+      .notification-dropdown {
+        max-width: 400px;
+      }
+    }
```
spike-rabbit: I am not really happy, that we show consumer here to use a media-query.

I think the whole purpose is to have a `max-width: 400px` in desktop mode, but `max-width: auto` in mobile, correct? I see basically two ways here:
1. we have a general max-width on header dropdowns
2. the header dropdown indicates with a class that it is open in mobile mode. Then you could write stlyes like this:
```css
.dropdown-menu.show-overlay {
  max-width: 400px;
}
```

PR #796, opened 2025-10-02
build(schematics): add schematics setup

spike-rabbit:
Sets up schematic infrastructure and adds
a hello-world generate schematic to test.

---


spliffone:
@spike-rabbit The changes look good, just two thoughts:
- with the ng-update `packageGroup` we are able to group and sync library versions. Means we could provide an extra schematics `@siemens/element-schematics` package and keep everything external including additional dependencies.  See [SAP Spartacus](https://github.com/SAP/spartacus/tree/develop/projects/schematics)
- Using the ng-packagr means we compile ESM but I believe Angular use CommonJS for compatibility reason. We should be careful choosing the output format since we might want to call external schematics in future.

kfenner (approved):
LGTM 👍

PR #799, opened 2025-10-03
feat(filter-bar): translatable aria label for remove icon

chintankavathia:
added `SiFilterBarComponent.removeFilterAriaLabel` input for providing translatable aria label text on remove icon.


spike-rabbit (changes requested):
I think the whole keyboard interaction is not ideal on this component. I think it should be a listbox (using the cdk component). Then this delete button can just be aria-hidden and we have a listener for the remove key.

WDYT?

spike-rabbit:
This PR is stale, and we will go with another solution

PR #815, opened 2025-10-07
feat(migration): add schematics for Siemens migration

mistrykaran91:
- added comprehensive import mappings for Simpl to Siemens component migration

**Testing:**

- Use local npm registry [verdaccio](https://verdaccio.org/docs/installation) 
- Do not forget to update the `registry` to `http://localhost:4873/` in `.npmrc` or `publishConfig` in your `package.json`
- update the element-ng version in `package.json`
- call `npm run build:all`
- change dir to `./dist/@siemens/element-ng`
- call `npm publish`
- call with example usage provided below

**Example usage without ng-add:**

```bash
ng generate @siemens/element-ng:simpl-siemens-migration
```
OR
```bash
ng generate @siemens/element-ng:s2s-migration
```

**Example usage with ng-add:**

```bash
npm run ng add @siemens/element-ng@<version> --force -- --force
```

---


projects/element-ng/schematics/collection.json
```diff
@@ -4,6 +4,16 @@
     "hello-world": {
       "description": "A blank schematic.",
       "factory": "./hello-world/index#helloWorld"
+    },
+    "siemens-migration": {
+      "description": "A migration schematic to update simpl imports to siemens imports.",
```
spike-rabbit, line 9:
```suggestion
      "description": "A migration schematic for repalcing @simpl/* packages with @siemens/*",
```

projects/element-ng/schematics/ng-add/index.ts
```diff
@@ -0,0 +1,19 @@
...
+import { getPackageJsonDependency } from '@schematics/angular/utility/dependencies';
+
+export const ngAdd = (options: { path: string }): Rule => {
+  return (tree: Tree, context: SchematicContext) => {
+    context.logger.info('🔧 Adding @siemens/element-ng to your project...');
+
+    const hasSimplElementNgDependency = getPackageJsonDependency(tree, '@simpl/element-ng');
+
+    if (hasSimplElementNgDependency) {
+      const chainedRules = chain([schematic('siemens-migration', options)]);
```
spike-rabbit, line 15: I think here should also be a prompt to the user.

I guess for actual migrations it might be nicer to first install.
Then commit changes and then run the migrations.

But the prompt we can do later. Just keept this so I can create an issue before merging.

projects/element-ng/schematics/siemens-migration/index.ts
```diff
@@ -0,0 +1,17 @@
...
+ * Copyright (c) Siemens 2016 - 2025
+ * SPDX-License-Identifier: MIT
+ */
+import { chain, Rule, SchematicContext, Tree } from '@angular-devkit/schematics';
+
+import { importMigrationRule } from './ts-import-migration-rule.js';
+
+// You don't have to export the function as default. You can also have more than one rule factory
+// per file.
+export const siemensMigration = (_options: { path: string }): Rule => {
```
spike-rabbit, line 11: Maybe we call the whole thing like:
- simplOpenSourceMigration
- simplSiemensMigration


siemensMigration is not pretty clear.

projects/element-ng/schematics/utils/project-utils.ts
```diff
@@ -0,0 +1,144 @@
...
+      if (tree.exists(tsConfig)) {
+        buildPaths.add(normalize(tsConfig));
+      }
+    }
+  }
+
+  return [...buildPaths];
+};
+
+export const createFullPathTree = (basePath: string, tree: Tree): Tree => {
```
spike-rabbit, line 46: It seems like we are doing some weird stuff here.

Please checkout out the angular approach:
This is how they parese the tsconfig: https://github.com/angular/angular/blob/main/packages/core/schematics/utils/typescript/parse_tsconfig.ts

They usually create a programm for the migration. I think we should follow that approach as well.
https://github.com/angular/angular/blob/main/packages/core/schematics/utils/typescript/compiler_host.ts

projects/element-ng/schematics/siemens-migration/index.ts
mistrykaran91: Done

projects/element-ng/schematics/ng-add/index.ts
mistrykaran91: Yes, in that case what we can do is run the dependency migration rule(which I'll create another PR) first and then we migrate things here.

projects/element-ng/schematics/utils/project-utils.ts
spliffone: @spike-rabbit we have to do it like this at the moment since we operate on the virtual tree. In case we add the following file [devkit_filesystem](https://github.com/angular/angular/blob/main/packages/core/schematics/utils/tsurge/helpers/angular_devkit/devkit_filesystem.ts) we can deal the typescript differently. The main issue typescript directly operate on the filesystem level and we can't since we need to use the virtual tree.

PR #825, opened 2025-10-08
refactor(typeahead): extract filtering into separate helper

### LINKED ISSUE #803: typeahead: extract search algorithm in seperate module
### Prerequisites

### Feature request

The search algorithm of the typeahead should be extracted into a new module, so that it is reusable.
I suggest an API / way of using similar to this:

```ts
class Search { // I think a better name would be nice
  constructor(datasource: SearchOptions[] | Observable<SearchOptions>| Signal<SearchOptions>)
  updateQuery(query: string): void;
  updateDatasource(datasource: ...): void;
  matches: Signal<Matches>;
}

interface SearchOptions<T> {
  // Text that is used for searching
  label: string;
  // Whatever data that should be linked to the text
  value: T;
}

interface Matches<T> {
  value: T;
  label: string;
  // Whatever we have currently in the typeahead
}
```

It then can be used like this:
```ts
@Component({
  <input (input)="input(inputValue)"
  {{ search.matches() | renderMeSomehow }}
})
class MyComponent {
  options = input()

  search = new Search(this.options);

  input(value: string) {
    this.search.updateQuery(value);
  }
}

```

### Use case

We have many upcoming requests in the filtered-search, which cannot be built with the typeahead.
In order to keep the typeahead clean it is best stop using it in the long run in the filtered-search.

Having this seperated would also allow us making use the search functionality in the select.

spike-rabbit:
This breaks down the code of the typeahead into smaller chunks. We achieve two goals by doing this:
- typeahead has increased maintainability by moving the logic into helper
- the helper is supposed to be reused directly in other components like FS and select

Closes #803

---


projects/element-ng/typeahead/si-typeahead.search.ts
```diff
@@ -0,0 +1,274 @@
+/**
+ * Copyright (c) Siemens 2016 - 2025
+ * SPDX-License-Identifier: MIT
+ */
+import { computed, Signal } from '@angular/core';
+
+export interface SearchOptions<T> {
+  /** The text that will be used for searching. */
```
timowolf: Specially here. I first thought this is the text, that the user enters. But no. The is the test that is used to match the search. It is constructed on the value of the option. I guess it could event be translated of a date object be transformed to a string.

PR #867, opened 2025-10-14
fix(filtered-search): scrollbar should appear only when needed

### LINKED ISSUE #847: Horizontal scroll-bar appears on si-filtered-search bar even though there is enough space available on search bar.
### Prerequisites

### Affected component

<si-filtered-search>

### Summary

There is a horizontal scroll bar present on si-filtered-search, even though there is enough space available on the search bar.
Refer attached GIF.

Image

### Version

_No response_

### Steps to reproduce

1. Open the live preview of SiMPL-element.
2. Select si-filtered-search/si-filtered-search-basic from examples.
3. Resize the screen, so that search bar gets almost 360px of width.
4. Then apply filter, we can see a **horizontal scroll bar appears** and the **input part where values needs to be entered takes a lot of space**, considering availability of space and requirement.

### Code reproduction example

https://element.siemens.io/element-examples/#/overview/si-filtered-search/si-filtered-search-basic?q=si-filtered

Horizontal scroll bar appears, even though enough space is available.

Horizontal scroll bar should appear only when there is no space to perform any operation.

akashsonune:
Fixes #847

### DIFF
--- a/projects/element-ng/filtered-search/si-filtered-search.component.html
+++ b/projects/element-ng/filtered-search/si-filtered-search.component.html
@@ -31,6 +31,7 @@
     type="text"
     class="value-input ps-2 me-2 flex-grow-1"
     typeaheadOptionField="translatedLabel"
+    [size]="searchValue.length || 1"
     [siTypeahead]="dataSource"
     [typeaheadMinLength]="readonly() ? 1 : 0"
     [typeaheadOptionsLimit]="typeaheadOptionsLimit()"


spike-rabbit (changes requested):
I like this approach, but this does not fully solve all issues:

Image

spike-rabbit:
I think the problem in my screenshot is actually, that the free-text-input is still consuming some space, although not being focues.

PR #872, opened 2025-10-15
docs(grid): document flexbox grid system

spliffone:

---


spike-rabbit:
Since we are copying documentation from BS, I think we need to adjust our license stuff.
This is the license of the BS 51. docs https://creativecommons.org/licenses/by/3.0/.
It required attribution, but as far as I can tell, but we are not having this yet.

@timowolf / @fh1ch any expert opinion here how to do that?

mkdocs.yml
```diff
@@ -44,6 +46,7 @@ nav:
       - Best practices:                     'fundamentals/ux-text-style-guide/best-practices.md'
       - Errors, warnings and notifications: 'fundamentals/ux-text-style-guide/error-messages.md'
       - Dialogs and buttons:                'fundamentals/ux-text-style-guide/dialogs-and-buttons.md'
+    - Grid system:                         'fundamentals/grid-system.md'
```
kfenner:
```suggestion
    - Grid system:                        'fundamentals/grid-system.md'
```

docs/fundamentals/layouts/grid.md
```diff
@@ -41,3 +40,193 @@ content to accommodate the browser window's width, regardless of how wide it
 becomes.
 
 ![Layout responsive](images/layout-fluid-fixed.png)
+
+## Code ---
+
+Leverage our mobile-first flexbox grid to create layouts of any complexity using a 12-column system, six responsive [breakpoints](./breakpoints.md), and a wide range of predefined utility classes.
```
kfenner: @spliffone  We already give credit to Bootstrap in the README but the grid system is probably one of the core concepts from Bootstrap that we should also give credit here directly - and it also helps people to understand the concept.

I don't think we need to repeat this for everything we derrive from Bootstrap, but here it just makes sense IMHO.

```suggestion
The flexbox grid is used to create layouts of any complexity using a 12-column system, six responsive [breakpoints](./breakpoints.md), and a wide range of predefined utility classes. The grid system is based on the well-known and beloved Bootstrap grid.
```

PR #891, opened 2025-10-20
fix(ip-input): shift remaining numbers to the right when inserting in middle of an ipv4 input

spliffone:
The input shift right the numbers to the remaining address parts

Explanation the pipe (|) represent the current cursor:
- `19|9.2.3` -> User presses `1` -> `191.|92.3`
- `19|2.168.0.1` -> User presses `1` -> `191.|216.80.1`
- `25|5.255.255.255` -> User presses `6` -> `25.6|5.255.255`

Behaviour changes:

- The IPv4 input no longer create a new address part by inserting a `.` when the user start with a 0. Now a user can type `000.000`, before a starting `0` was converted to `0.`
- Removing the last number in an address part automatically dropped the `.` this behaviour was removed

---

### DIFF
--- a/projects/element-ng/ip-input/address-utils.ts
+++ b/projects/element-ng/ip-input/address-utils.ts
@@ -69,7 +69,7 @@ const recursiveSplitIpV4 = (
         cursorDelta = 1;
       }
     }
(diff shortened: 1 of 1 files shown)


spike-rabbit (approved):
👍

PR #900, opened 2025-10-22
docs(position): document position utils

spliffone:

---


kfenner:
@spliffone Looks good in general. Can you either provide a more realistic example, or maybe just delete it? 

image

I think this is not really needed, the description and examles above should be enough to understand how to use it, don't you think?

kfenner (approved):
LGTM 👍

PR #901, opened 2025-10-22
feat(header-dropdown): add class `header-dropdown-overlay` when rendered as overlay

ljanner:
Adds `header-dropdown-overlay` class based on rendering mode, enabling distinct styling between inline and overlay dropdown modes.

Can be used in cases like #791

---


spike-rabbit (changes requested):
Please label the `@internal` commit as `docs(): ...`. I don't think it needs to show up in the changelog.

Code is good already 👍

ljanner:
> Please label the `@internal` commit as `docs(): ...`. I don't think it needs to show up in the changelog.

Done ✔️

PR #906, opened 2025-10-24
feat(resize-observer): provide `borderBoxSize` and `contentBoxSize` o…

spliffone:
…n resize

---

### DIFF
--- a/projects/element-ng/resize-observer/si-resize-subscriptions.ts
+++ b/projects/element-ng/resize-observer/si-resize-subscriptions.ts
@@ -110,6 +110,16 @@ export class SiResizeSubscriptions {
   }
 
   private handleElement(event: ResizeEvent): void {
+    // eslint-disable-next-line no-console
+    console.log(
+      'resize event:',
+      this.box,
+      'borderBox',
+      event.borderBoxSize?.at(0),
+      'contentBox',
+      event.contentBoxSize?.at(0),
+      event.target
+    );
     const element = event.target;
     const entry = this.listeners.get(element);
     if (!entry) {
--- a/projects/element-ng/resize-observer/si-resize-observer-utils.ts
+++ b/projects/element-ng/resize-observer/si-resize-observer-utils.ts
@@ -12,6 +12,8 @@ export const getElementSize = (
 ): { width: number; height: number } => {
   let width = element.clientWidth;
   let height = element.clientHeight;
+  // eslint-disable-next-line no-console
+  console.log('client width', width, 'height', height);
   if (box === 'border-box') {
     // The border box calculation is based on https://github.com/juggle/resize-observer
     const cs = getComputedStyle(element);
--- a/src/app/examples/resize-observer/resize-observer.html
+++ b/src/app/examples/resize-observer/resize-observer.html
@@ -1,0 +1,8 @@
+<div #main class="border-1 m-5">
+  @for (item of items(); track $index) {
+    <button appDummy type="button" class="btn btn-secondary m-1">
+      {{ item }}
+    </button>
+  }
+</div>
+<button type="button" class="btn" (click)="changeSize()">Change size</button>
--- a/src/app/examples/resize-observer/resize-observer.ts
+++ b/src/app/examples/resize-observer/resize-observer.ts
@@ -1,0 +1,64 @@
+/**
+ * Copyright (c) Siemens 2016 - 2025
+ * SPDX-License-Identifier: MIT
+ */
+import {
+  Component,
+  contentChild,
+  contentChildren,
+  Directive,
+  effect,
+  ElementRef,
+  inject,
+  OnInit,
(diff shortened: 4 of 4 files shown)


projects/element-ng/resize-observer/resize-observer.service.ts
```diff
@@ -118,27 +129,38 @@ export class ResizeObserverService {
...
 
-  private handleResizeSubscriber(element: Element, entry: ResizeSubscriber): void {
+  private handleResizeSubscriber(event: ResizeEvent, entry: ResizeSubscriber): void {
     if (entry.blocked) {
       return;
     }
+    const element = event.target;
+    if (entry.dim) {
+      entry.dim.borderBoxSize = event.borderBoxSize?.at(0);
+      entry.dim.contentBoxSize = event.contentBoxSize?.at(0);
```
kfenner: Did you test that you're getting a proper value here? I thought @dr-itz said you can only monitor one (border-box or content-box). See also the `ResizeObserver.observe()` parameters where you can specify the option: 
`type ResizeObserverBoxOptions = "border-box" | "content-box" | "device-pixel-content-box";`

kfenner: And on top, I think we said in the internal discussion, it would also not make sense to monitor both for the same element.

projects/element-ng/resize-observer/resize-observer.service.ts
```diff
@@ -9,6 +9,8 @@ import { Observable, Subscriber } from 'rxjs';
 export interface ElementDimensions {
   width: number;
   height: number;
+  borderBoxSize?: ResizeObserverSize;
+  contentBoxSize?: ResizeObserverSize;
```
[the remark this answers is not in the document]
kfenner: I'm fine with this, but I think we should add some docs comments to quickly explain what inline/block  size are, would be really helpful, no?

projects/element-ng/resize-observer/resize-observer.service.ts
dr-itz: monitor for what the consumer is asking, i.e. border-box by default.

- create two ResizeObserver, one for border-box, one for content-box
- handling between those is independent
- register with the one that is needed based on consumer option

i.e. I would move most of the current service to a helper class and have two instances of those around, i.e. make the current service a thin mux layer

api-goldens/element-ng/resize-observer/index.api.md
```diff
@@ -26,20 +26,17 @@ export interface Breakpoints {
...
     width: number;
 }
 
 // @public
 export class ResizeObserverService {
-    constructor();
     // @deprecated
     _checkAll(): void;
-    observe(element: Element, throttle: number, emitInitial?: boolean, emitImmediate?: boolean): Observable<ElementDimensions>;
+    observe(element: Element, throttle: number, emitInitial?: boolean, emitImmediate?: boolean, box?: 'content-box' | 'border-box'): Observable<ElementDimensions>;
```
spike-rabbit: I think the parameter list got too long.

Can you just add an overload like this:

```suggestion
    observe(element: Element, options: { throttle: number, emitInitial?: boolean, emitImmediate?: boolean, box?: 'content-box' | 'border-box'}): Observable<ElementDimensions>;
```

spliffone: Changing the signature parameter signature is breaking, I guess we should provide a second function where we can use the options. Any name suggestions?

projects/element-ng/resize-observer/si-resize-subscriptions.ts
```diff
@@ -0,0 +1,207 @@
...
+
+  /**
+   * @internal
+   */
+  checkAll(): void {
+    this.listeners.forEach(entry => this.handleElement({ target: entry.element }));
+  }
+
+  private emitSize(element: Element, entry: ResizeSubscriber, force = false): void {
+    const dimensions = getElementSize(element, this.box);
```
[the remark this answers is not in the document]
spike-rabbit: But the content-box seems to be also the size without scrollbar. I just check it.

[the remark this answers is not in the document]
spliffone: @spike-rabbit I just checked the implementation against a button and @dr-itz is right I don't provide the correct result regarding the resize-observer implemenation:
1. The client width/height does not include the border
2. The resize observer content-box provide pure content size without padding and border
3.  The resize observer border-box provide the size with padding and border

What is your opinion, should I adjust the code to handle 3 cases: client-size, content-box, border-box?

spliffone:
I am closing the PR since the required changes to provide the `contentBox` and `borderBox` to the listeners is by far to complex. Since the `ResizeObserverService` provide a functionality to initially emit the size we have to calculate the boxes internally.

PR #909, opened 2025-10-24
ci: use bun

Killusions:
Experimenting with install and build time differences

---


spike-rabbit:
Currently no plans to switch to bun.

PR #910, opened 2025-10-24
docs(layouts): update docs

panch1739:
This PR updates the docs for the layout chapter, to improve the ux designer content.

---


panch1739:
@kfenner 

- I tried to normalize the structure of the chapters by adding usage/design/code tab when possible. Im not sure if is necesary
- I did not added new images...i feel that the new content is quite technical and the parts that are valid for designers, do not need images. E.g., the table with the breakpoints
- I renamed the Grid chapter to "Flexbox grid" and move the design part to the overview...i think it makes the most sense, as we just provide a generality of the layout and how it works. In there, i renamed the section to "Design grid". 
- I could improve more the design part, but maybe for now is ok to have it cleaner.

Let me know your thoughts! 

@spliffone I have written to you privately with some questions, maybe we can check together when you have time :)

docs/fundamentals/layouts/gutters.md
```diff
@@ -1,13 +1,16 @@
...
-Gutters are the paddings between columns in the [grid system](./grid.md#code), providing responsive separation and alignment for your content.
+Gutters are the paddings between columns in the [grid system](./grid.md#code),
+providing responsive separation and alignment for the content.
 
-## Concept
+## Usage ---
 
 - Gutters refer to the horizontal gaps between column content, achieved by applying padding to the start and end of each column. Negative margins on the row ensure content remains properly aligned at the edges.
-- By default, gutters are `1.5rem` (`24px`) wide, matching the grid’s spacing scale for consistent layout.
+- By default, gutters are `1.rem` (`16px`) wide, matching the grid’s spacing scale for consistent layout.
```
spliffone:
```suggestion
- By default, gutters are `1rem` (`16px`) wide, matching the grid’s spacing scale for consistent layout.
```

docs/fundamentals/layouts/gutters.md
```diff
@@ -1,13 +1,16 @@
...
 - Gutters can be adjusted responsively using breakpoint-specific classes, allowing you to control horizontal, vertical, or both types of spacing as needed.
 
+## Code ---
```
spliffone: The Code section has the same level like the sub chapters `Horizontal gutters`, ..., is this by purpose?

docs/fundamentals/layouts/grid.md
```diff
@@ -1,52 +1,9 @@
-# Grid
+# Flexbox grid
 
-**Layout principles** allow the arrangement of elements in an interface.
-They help developers and designers to ensure consistency, efficiency,
-and effective communication in the user experience.
+The **flexbox grid** is used to create layouts of any complexity using a 12-column system, six responsive [breakpoints](./breakpoints.md), and a wide range of predefined utility classes. The grid system is based on the well-known and beloved Bootstrap grid.
 
 ## Usage ---
```
kfenner: @panch1739 Since we are now left with a developer only docs here in this chapter, we could get rid of the `usage` and `code` tabs and instead have a better right side navigation tree.

mkdocs.yml
```diff
@@ -24,13 +24,13 @@ nav:
...
       - Overview:                         'fundamentals/layouts/overview.md'
-      - Anatomy:                          'fundamentals/layouts/anatomy.md'
-      - Header:                           'fundamentals/layouts/header.md'
-      - Content:                          'fundamentals/layouts/content.md'
       - Breakpoints:                      'fundamentals/layouts/breakpoints.md'
-      - Grid:                             'fundamentals/layouts/grid.md'
+      - Flexbox grid:                     'fundamentals/layouts/grid.md'
       - Gutters:                          'fundamentals/layouts/gutters.md'
       - Spacing:                          'fundamentals/layouts/spacing.md'
+      - Anatomy:                          'fundamentals/layouts/anatomy.md'
```
kfenner: @panch1739  Not sure if we should put the highlevel chapters (anatomy, header, content) at the bottom? I think they are the more interesting ones and should remain at the top and the grid with all the details like gutters etc. should be at the bottom. The grid is more fundamental but not really required to read for a designer and acts more as a reference for devs when they implement something.

kfenner (changes requested):
@panch1739 Thank you so much for this PR. Moving the design view to the overview chapter is great IMHO. 
I just added a few comments - I think I'd get rid of the usage/code/desing tabs for these chapters as the docs have more of a developer viewpoint and it's a bit strange to break this up into these tabs. If we need to add a UX designer viewpoint, then it makes sense, but I don't think it's much help if we write about `sm-` classes in the usage chapter for designers. I understand that you did not optimize on that - but my question is, will we even have to add anything for designers here?

mkdocs.yml
kfenner: @panch1739 Thanks for the adaptations 🙏 
Let's go with your suggestion for now until we have all the other chapters in and then we can adjust when needed.

PR #915, opened 2025-10-27
feat: add experimental markdown renderer

Killusions:
Adds an experimental Markdown formatter and common markdown styles (applied via class).

---

### DIFF
--- a/playwright/e2e/element-examples/static.spec.ts
+++ b/playwright/e2e/element-examples/static.spec.ts
@@ -106,3 +106,5 @@ test('typography/color-variants', ({ si }) => si.static());
 test('typography/type-styles', ({ si }) => si.static());
 test('typography/display-styles', ({ si }) => si.static());
 test('typography/typography', ({ si }) => si.static());
+test('si-markdown-renderer/si-markdown-renderer', ({ si }) =>
+  si.static({ disabledA11yRules: ['link-in-text-block'] }));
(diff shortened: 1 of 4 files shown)


src/app/examples/si-text-formatters/si-markdown-content.ts
```diff
@@ -0,0 +1,98 @@
...
+import { Component } from '@angular/core';
+import { SiMarkdownContentComponent } from '@siemens/element-ng/text-formatters';
+
+@Component({
+  selector: 'app-sample',
+  imports: [SiMarkdownContentComponent],
+  templateUrl: './si-markdown-content.html'
+})
+export class SampleComponent {
+  markdownContent = `# AI Assistant Response
```
kfenner: instead of inlining the markdown text, wouldn't it be nicer to just load it as example-text.md file?

projects/element-theme/src/styles/components/_markdown.scss
```diff
@@ -0,0 +1,127 @@
...
+    padding-inline-start: map.get(variables.$spacers, 4);
+
+    p {
+      margin-block-end: 0;
+    }
+  }
+
+  > *:first-child {
+    // stylelint-disable-next-line declaration-no-important
+    margin-block-start: 0 !important;
```
kfenner: why all the `important`? no better way to generally increase specificity?

projects/element-ng/text-formatters/markdown-formatter.ts
```diff
@@ -0,0 +1,242 @@
...
+ * - Returns a DOM node containing the formatted content.
+ *
+ * **Warning:** The returned Node is inserted without additional sanitization.
+ * Input content is sanitized before processing.
+ *
+ * @experimental
+ * @param sanitizer - Angular DomSanitizer instance
+ * @returns A function taking the markdown content to transform and returning a DOM div element containing the formatted HTML
+ */
+export const getMarkdownFormatter = (sanitizer: DomSanitizer): ((content: string) => Node) => {
```
kfenner: Not sure whether we should call this `renderer` or `formatter`. For me formatter sounds a bit more simpler, like formatting a number or a date string. This here feels more like a renderer to me, so I suggest to rename it but no strong opinion.

kfenner (commented):
I really like this, brilliant work 👏 🚀 
Left a couple minor comments.

kfenner (changes requested):

PR #916, opened 2025-10-27
feat: add si-action-bar for internal reuse

Killusions:
Adds experimental si-action-bar component with a purposefully minimal API (Interface) for internal reuse in the chat-components and also the notification-item (or named similarly)

---


kfenner (commented):
Not sure about this one, feels like a major overlap with the new button group variations: https://github.com/siemens/element/pull/588

If this is really something different, fine, but then I'd like to know the exact details, and also, why we cannot solve it using the button group? I think @dauriamarco is finishing this up so could be ready soon.

Killusions:
@kfenner Apart from containing buttons which are next to each other, this is not very similar to the button group, firstly, these are round (currently not possible in the button group), have a gap and are ghost buttons, but more importantly, the main reason for this component is to handle the action service and the dropdown menu, as well as providing a restrictive interface (a lot more restrictive than MenuItem) that we can re-expose for other components. Alone in the chat components, this exact code would have to be repeated 3 times (user chat, ai chat, chat input, possibly even more in the future), with Loris' notification item (or whatever it's called) also implementing basically the same.

PR #917, opened 2025-10-27
feat(resize-observer): provide a resize observer signal

spliffone:

---

/cc @siemens/siemens-element Can you please provide some feedback to the naming? Maybe you have a better name for the signal `observeElementSize`.

---

[Documentation](https://d33c9dcnqinn2a.cloudfront.net/pr-917/pages/).
[Examples](https://d33c9dcnqinn2a.cloudfront.net/pr-917/pages/element-examples/).
[Dashboards Demo](https://d33c9dcnqinn2a.cloudfront.net/pr-917/pages/dashboards-demo/).
[Playwright report](https://d33c9dcnqinn2a.cloudfront.net/pr-917/playwright-report/).

**Coverage Reports:**

Code Coverage

- [charts-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-917/pages/coverage/charts-ng/)
- [dashboards-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-917/pages/coverage/dashboards-ng/)
- [element-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-917/pages/coverage/element-ng/)
- [element-translate-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-917/pages/coverage/element-translate-ng/)
- [maps-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-917/pages/coverage/maps-ng/)
- [native-charts-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-917/pages/coverage/native-charts-ng/)


projects/element-ng/resize-observer/si-observe-element-size.ts
```diff
@@ -0,0 +1,79 @@
...
+
+  const [get, set] = createSignal<ReturnType>(mapFn([]));
+  const observer = new ResizeObserver(entries => set(mapFn(entries)));
+  inject(DestroyRef).onDestroy(() => observer.disconnect());
+
+  if (isSignal(source)) {
+    let currentEl: undefined | Element | ElementRef;
+    effect(() => {
+      const nextEl = source();
+      if (nextEl && nextEl !== currentEl) {
```
spike-rabbit: I think when nextEl is undefined you need still stop observing the currentEl

projects/element-ng/resize-observer/si-observe-element-size.ts
```diff
@@ -0,0 +1,79 @@
...
+  source: Source,
+  opt?: ResizeSignalOptions<ReturnType>
+): Signal<ReturnType> => {
+  const mapFn =
+    opt?.mapFn ?? ((entries: ResizeObserverEntry[]) => entries as unknown as ReturnType);
+  if (!isPlatformBrowser(inject(PLATFORM_ID)) || typeof ResizeObserver === 'undefined') {
+    return signal(mapFn([]));
+  }
+
+  const [get, set] = createSignal<ReturnType>(mapFn([]));
```
spike-rabbit: why not just using the normal `signal` function and just return it as readonly in the end?

projects/element-ng/resize-observer/si-observe-element-size.ts
```diff
@@ -0,0 +1,79 @@
...
+  Signal
+} from '@angular/core';
+import { createSignal } from '@angular/core/primitives/signals';
+
+/** Options for the resize signal. */
+export interface ResizeSignalOptions<ReturnType = ResizeObserverEntry[]> {
+  /** Sets which box model the observer uses. */
+  box?: ResizeObserverBoxOptions;
+  /** Custom to mapping function to transform value to the desired type. */
+  mapFn?: (entries: ResizeObserverEntry[]) => ReturnType;
```
spike-rabbit: I am not fully convinced we need this.
I think for the user the most convenient is, if we just return `ResizeObserverEntry`, not even the array.
Mapping on our end just makes the type here complicated. 
If conversion is needed, a `computed` block works just fine

spike-rabbit (changes requested):

projects/element-ng/resize-observer/si-observe-element-size.ts
```diff
@@ -0,0 +1,93 @@
...
+  opt?: ResizeSignalOptions
+): Signal<ResizeObserverEntry[]> => {
+  if (!isPlatformBrowser(inject(PLATFORM_ID)) || typeof ResizeObserver === 'undefined') {
+    return signal([]).asReadonly();
+  }
+
+  const options = { box: opt?.box ?? 'content-box' };
+  const sizeSignal = signal<ResizeObserverEntry[]>([]);
+  const observer = new ResizeObserver(entries => sizeSignal.set(entries));
+  inject(DestroyRef).onDestroy(() => observer.disconnect());
```
spike-rabbit, line 69: an effect has a clean-up function. I guess using this would be more reliable: https://angular.dev/guide/signals/effect#effect-cleanup-functions

projects/element-ng/resize-observer/si-observe-element-size.ts
```diff
@@ -0,0 +1,93 @@
...
+  ElementRef,
+  inject,
+  isSignal,
+  PLATFORM_ID,
+  signal,
+  Signal
+} from '@angular/core';
+
+/** Options for the resize signal. */
+export interface ResizeSignalOptions {
```
spike-rabbit:
```suggestion
export interface ObserverOptions {
```

I don't think there is a need to have `signal` in the name.

projects/element-ng/resize-observer/si-observe-element-size.ts
```diff
@@ -0,0 +1,93 @@
...
+  /** Sets which box model the observer uses. */
+  box?: ResizeObserverBoxOptions;
+}
```
spike-rabbit, line 21: What about allowing to pass parts of [CreateEffectOptions](https://angular.dev/api/core/CreateEffectOptions)?
Injector make sense.
DebugName why not?
I think manual cleanup not, as we do not expose the EffectRef

spike-rabbit (changes requested):

projects/element-ng/resize-observer/si-observe-element-size.ts
spliffone: That is the reason I use an option object we might extend the interface in future but for now I can't see a need for it

projects/element-ng/resize-observer/si-observe-element-size.ts
spliffone: That is true but the effect is only in use if the user provide a signal, the else case still need the destroyRef

spike-rabbit: You could just move the else part also inside the effect. I guess this may also improves readability.

projects/element-ng/auto-collapsable-list/si-auto-collapsable-list.directive.ts
```diff
@@ -68,114 +87,91 @@ export class SiAutoCollapsableListDirective implements AfterViewInit, OnChanges,
...
   /**
    * The same as {@link gap}, but automatically read from the computed styles.
    * Used if not set by user.
    */
   private computedGap = 0;
-  private injector = inject(INJECTOR);
+
+  constructor() {
+    this.destroyRef.onDestroy(() => this.disableInitSubscription?.unsubscribe());
+    effect(() => {
```
spike-rabbit, line 112: I think you can use computed here and get rid the effect entirely.

spike-rabbit (changes requested):
the resizeObserver looks good to me, maybe we should just discuss the name in the team.

projects/element-ng/auto-collapsable-list/si-auto-collapsable-list.directive.ts
spliffone: I can convert the first part to an computed but the second part still require an effect since we modify properties. But I have the feeling it get less readable and clear if we split the two parts.

projects/element-ng/auto-collapsable-list/index.ts
```diff
@@ -5,6 +5,5 @@
 export * from './si-auto-collapsable-list-item.directive';
 export * from './si-auto-collapsable-list.directive';
 export * from './si-auto-collapsable-list-overflow-item.directive';
-export * from './si-auto-collapsable-list-measurable.class';
```
chintankavathia, line 8: this would be a breaking change technically

projects/element-ng/resize-observer/si-observe-element-size.ts
```diff
@@ -0,0 +1,79 @@
...
+ * @param source - An `Element`, `ElementRef`, or a signal resolving to one.
+ * @param opt - Optional configuration for the underlying `ResizeObserver`.
+ * @returns A read-only signal emitting the latest `ResizeObserverEntry`.
+ * @remarks Must be called within an injection context (e.g. constructor,
+ * field initializer, or `runInInjectionContext`).
+ */
+export const observeElementSize = (
+  source: Source,
+  opt?: ObserveElementSizeOptions
+): Signal<ResizeObserverEntry | undefined> => {
```
chintankavathia: since the method is exported, its param type should also be exported. I would use more semantic name like `ObserveElementSizeSource` instead of `Source`

chintankavathia (changes requested):

projects/element-ng/resize-observer/si-observe-element-size.ts
```diff
@@ -0,0 +1,79 @@
...
+ * The observer is automatically disconnected when the enclosing
+ * `DestroyRef` is destroyed.
+ *
+ * @param source - An `Element`, `ElementRef`, or a signal resolving to one.
+ * @param opt - Optional configuration for the underlying `ResizeObserver`.
+ * @returns A read-only signal emitting the latest `ResizeObserverEntry`.
+ * @remarks Must be called within an injection context (e.g. constructor,
+ * field initializer, or `runInInjectionContext`).
+ */
+export const observeElementSize = (
```
chintankavathia: should also mark this as experimental

kfenner:
We mark it as experimental, so if we figure out a better name later, we can adjust it.

chintankavathia (approved):
LGTM

kfenner (commented):
Solid direction. The new `observeElementSize` is well documented, the spec is genuinely good (real browser, correct double-rAF wait, covers all four source shapes + destroy + box option), and the `auto-collapsable-list` refactor removes a lot of RxJS plumbing while fixing a real bug (content-box now comes from the observer instead of manual padding subtraction). Findings below, roughly by severity.

## Blocking

**1. One `ResizeObserver` instance per call — contradicts the documented architecture in the same entry point**

`resize-observer.service.ts` states *"only one `ResizeObserver` should be used for performance reason."* `observeElementSize` allocates a fresh one per invocation, and `SiAutoCollapsableListMeasurable` calls it per item. A content action bar with 4 actions now creates ~6 observers where it previously shared 1 — every list, breadcrumb, tabset and filter bar multiplies this. Suggest backing the signal with a root-provided registry (`Map<Element, Set<sink>>`, keyed by box) that owns a single observer and fans entries out to per-call signals. The public API stays identical.

**2. `MockResizeObserver.instances` is never cleared**

Neither `mockResizeObserver()` nor `restoreResizeObserver()` resets the static array, so it grows for the lifetime of the test process and `triggerResize()` fans out to observers belonging to already-destroyed fixtures. Reset it in `mockResizeObserver()`. (It's not a published entry point, so the `instance` → `instances` rename itself is fine.)

## Should address

**3. Docs not updated.** `docs/architecture/resize-observer.md` still presents only the service and directive approaches. The new public API needs a section plus guidance on which to prefer going forward.

**4. Returning the raw `ResizeObserverEntry` is easy to misuse.** Both call sites write `entry()?.contentBoxSize[0].inlineSize` / `borderBoxSize[0].inlineSize`. The `box` option and the property read are fully decoupled — `{ box: 'border-box' }` + reading `contentBoxSize` type-checks and silently measures the wrong thing. Consider resolving against the configured box and returning `Signal<{ inlineSize; blockSize } | undefined>`. Also `contentBoxSize[0]` is unchecked; the spec permits an empty array for fragmented/exotic targets, and `source` accepts any `Element`.

**5. `readGapSize()` source changed** from `containerElement() ?? host` to always `host` in `si-auto-collapsable-list.directive.ts`. Looks like an intended fix, but it's an unannounced behavior change — worth a line in the commit body.

**6. `computedGap` is non-reactive but read inside the effect.** It's a plain field set in `ngAfterViewInit`, while the effect is created in the constructor. It only works because the effect bails until the first `ResizeObserver` callback lands. Making it a `signal` set from `afterNextRender` would make the dependency explicit rather than incidental.

## Nits

- Dropping `SiAutoCollapsableListMeasurable` from `index.ts` is technically a public API removal, but it was `@public (undocumented)`, has no usages in `src/` or `docs/`, and its only member was `size$` — really an internal measurement channel. Fine to leave unflagged. What does look off is that the regenerated golden still has `extends SiAutoCollapsableListMeasurable` in two places while no longer declaring the type; api-extractor normally emits `ae-forgotten-export` for that, so it may be worth checking whether that rule is suppressed.
- Typos in the new comment: `wo  would latch` → `we would latch` (double space too), `Size the items influence` → `Since the items influence`.
- `@experimental` is on `observeElementSize` and `ObserveElementSizeSource` but missing on `ObserveElementSizeOptions`.
- `opt` → `options` (param and `@param`).
- `SourceElement` duplicates the non-signal half of `ObserveElementSizeSource`; could reuse it: `export type ObserveElementSizeSource = SourceElement | Signal<SourceElement | undefined | null>;`
- The doc says "automatically **disconnected** when the enclosing `DestroyRef` is destroyed", but cleanup only calls `unobserve` — `disconnect()` is never called. Either reword or add `DestroyRef.onDestroy(() => observer.disconnect())`.
- `entries.at(-1)` hides the "only ever one observed element" assumption; `entries[0]` plus a short comment reads clearer.
- The `chat-messages` golden change looks like unrelated regen drift — worth calling out so reviewers don't hunt for a missing source change.
- The directive now mixes styles: `disableInitSubscription` + `toObservable` alongside the new effect. Could be one more effect.

## Missing test coverage

- No case for the non-browser / `ResizeObserver === undefined` branch (permanently-`undefined` signal).
- Nothing covers `SiContentActionBarComponent`'s `afterNextRender` parent-element change.
- Directive coverage is otherwise fine — `useContainerElement` and disable/re-enable are both exercised.

## On the naming question

`observeElementSize` reads like an imperative that returns a subscription, and it doesn't actually return a *size*. Options:

| Name | Fit |
| --- | --- |
| `elementSize` | Cleanest, matches the `linkedSignal` / `httpResource` noun style — but only honest if you adopt #4 and return a size |
| `elementResize` / `resizeEntry` | Accurate if the return stays `ResizeObserverEntry` |
| `injectElementSize` | Makes the injection-context requirement obvious; common community convention |

My preference: adopt #4 and call it **`elementSize`**. If the entry stays as-is, **`elementResize`** fits better than the current name.

---

<sub>_Edited: downgraded the `SiAutoCollapsableListMeasurable` removal from blocking to a nit — it was undocumented and effectively unused, so a `BREAKING CHANGE` footer isn't warranted. Remaining items renumbered._</sub>

kfenner (commented):
## Re-review at `8a5a8a8`

Strong revision — nearly everything from the previous pass landed, and the registry went further than I suggested.

### Resolved

| Was | Now |
| --- | --- |
| Per-call `ResizeObserver` (blocking) | `ResizeObserverRegistry` — and `ResizeObserverService` was refactored onto it too, so the whole library shares **one** native observer. Better than what I asked for. |
| `MockResizeObserver.instances` never cleared (blocking) | Reset in `mockResizeObserver()`, plus a dedicated `resize-observer.mock.spec.ts` |
| Docs not updated | Good new section with an explicit preference order and a "don't fall back to `clientWidth`" warning |
| Raw `ResizeObserverEntry` easy to misuse | Returns `ElementSize`, resolved against the configured box, `.at(0)` + `size && …` guard, and a spec for the empty-fragment case |
| `readGapSize()` unannounced change | Now covered by `should read the gap from the host when using an external container` |
| `computedGap` non-reactive | `signal(0)` set from `afterNextRender` |
| Typos, "disconnected" wording, `entries.at(-1)` | All fixed |
| Naming | `elementSizeSignal` — good call, it's honest now that it returns a size rather than an entry |

### New findings in the registry

**1. Multiple box options on the same element silently degrade.**

`getObservedBox` picks a **single** box by fixed priority (`content-box` > `border-box` > `device-pixel-content-box`), but `emit` fans the entry out to *all* subscribers for that element regardless of which box they asked for.

The reported values stay correct — a `ResizeObserverEntry` always carries all three size arrays — but the **notification trigger** is wrong. A `border-box` subscriber on an element that also has a `content-box` subscriber only wakes on content-box changes, so a padding change on a fixed content-width element never fires. Same for `device-pixel-content-box` and a DPR/zoom change.

Not reachable in this repo today — I checked every `siAutoCollapsableList`, `siResizeObserver` and `siResponsiveContainer` host and no element is observed twice with different boxes. But `elementSizeSignal` is public API and `ResizeObserverService` now hard-codes `'content-box'`, so a consumer putting `elementSizeSignal(el, { box: 'border-box' })` on an element that also carries `(siResizeObserver)` gets a quietly stale signal.

Suggestion: hold one native observer **per box option** — `Map<ResizeObserverBoxOptions, ResizeObserver>`, so at most three for the whole app. Still shared and bounded, and it removes both `getObservedBox` and the `refresh` unobserve/observe dance. Either way there is currently no test for this case.

**2. `ElementSize` identity churn defeats signal equality.**

`getBoxSize` returns a fresh object literal on every emit, so with the default `Object.is` equality the signal notifies even when the dimensions are unchanged. This is worth adding now that the value is plain data — it wasn't practical with the raw entry:

```ts
const sizeSignal = signal<ElementSize | undefined>(undefined, {
  equal: (a, b) => a?.inlineSize === b?.inlineSize && a?.blockSize === b?.blockSize
});
```

**3. Each new subscriber re-notifies every existing subscriber on that element.**

`observe()` passes `refresh = true`, forcing `unobserve` + `observe` so the newcomer receives an initial callback. That callback fires for the target and `emit` then fans it out to all sinks. The intent is right, but combined with (2) every already-registered subscriber sees a spurious change. Fixing (2) makes this harmless.

Also, the explicit `unobserve` before `observe` is redundant: per spec, `observe()` already replaces an existing observation for the same target.

**4. `ResizeObserverRegistry` has no `ngOnDestroy`.**

Irrelevant in production since it is root-provided and lives for the app lifetime, but every `TestBed.resetTestingModule()` builds a new registry and leaks the previous native observer. One-liner:

```ts
ngOnDestroy(): void {
  this.resizeObserver?.disconnect();
}
```

**5. Question:** the registry wraps every emit in `zone.run()`. For the signal path that shouldn't be necessary — signal writes schedule their own notification — and in a zone-based app it forces a change detection cycle per resize batch. Was this needed specifically for the `ResizeObserverService` path?

### Still open from the previous round

- `@experimental` missing on `ElementSizeSignalOptions`; the other three exports have it.
- `opt` → `options` (both the parameter and the `@param` tag).
- `SourceElement` is still duplicated inside `ElementSizeSignalSource`.
- `disableInitSubscription` + `toObservable` still sit alongside the effect in the auto-collapsable-list directive.
- The `chat-messages` golden still carries the unrelated `+import { Signal }` drift.
- The auto-collapsable-list golden still has `extends SiAutoCollapsableListMeasurable` in two places without declaring the type.

### Test gaps

- No case for the non-browser / `ResizeObserver === undefined` branch.
- Nothing covers the `afterNextRender` change in `SiContentActionBarComponent`.
- No case for two subscribers on one element with different box options (see finding 1).

projects/element-ng/resize-observer/testing/resize-observer.mock.ts
```diff
@@ -45,14 +53,24 @@ export class MockResizeObserver {
...
-  this.observed.push([target, options])
-  )
+   * vi.fn((target: Element, options?: ResizeObserverOptions) => {
+   *   const index = this.observed.findIndex(([element]) => element === target);
+   *   if (index === -1) {
+   *     this.observed.push([target, options]);
+   *   } else {
+   *     this.observed[index] = [target, options];
+   *   }
+   * })
```
dr-itz: can we instead just disable the linter rule that is completely useless here?

dr-itz:
Frankly, I think we should not modify the resize-observer service at all here. It works and is proven to do so. Using this new registry is a refactoring that _might_ be done later. And I'm not sure it's worth the added complexity.

What remains fundamentally unsolved is the fact that a single element cannot be observed for different box handling. In reality this is very unlikely to matter at all.

dr-itz:
Another thing, and I mentioned this multiple times during calls, is that the signal does not support throttling or batched updates like the service. And this is crucial for things like resizing charts. So I wonder if it wouldn't be much simpler to have the signal backed by the service instead of yet another service

kfenner:
@spliffone @spike-rabbit @dr-itz @chintankavathia Discussed this in the stand-up and decided to not merge because there are still concerns. The "breaking change" is so minor, we could just merge it in a normal 51.x release in my opinion.

projects/element-ng/resize-observer/si-element-size-signal.ts
```diff
@@ -0,0 +1,125 @@
...
+};
+
+const equalElementSize = (
+  previous: ElementSize | undefined,
+  current: ElementSize | undefined
+): boolean =>
+  previous === current ||
+  (!!previous &&
+    !!current &&
+    previous.inlineSize === current.inlineSize &&
```
spliffone, line 103: I guess if we want to got with a equal function we should verify the absolute values and ignore half pixel shifts like `inline-size: 20.12`. 
@dr-itz  @spike-rabbit  @chintankavathia What do you think?

PR #927, opened 2025-10-29
test: textarea scrollbar

ljanner:
Scrollbars will now be visible in the e2e test snapshots, which makes it easier to spot overlapping scenarios like the severity icon in the text-area.

---

### DIFF
--- a/playwright.config.ts
+++ b/playwright.config.ts
@@ -55,7 +55,8 @@ const chromeLaunchOptions = {
     '--disable-oop-rasterization',
     '--disable-composited-antialiasing',
     '--disable-smooth-scrolling'
-  ]
+  ],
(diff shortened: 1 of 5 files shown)


ljanner:
@spike-rabbit @Killusions What do you guys think about adding the `ignoreDefaultArgs: ['--hide-scrollbars']` to the playwright config?

It's a default arg for chrome headless and to my understanding can't be changes test specific therefore the arg has to be ignored globally.

As shown in the playwright report this has many changes for other snapshots. Snapshots like `si-landing-page/si-landing-page-basic-login` or the chart static snapshots stand out - with a correct viewport we can avoid scrollbars in these snapshots.

Should we continue with the adaptions to the playwright config and adjust VRTs like static charts or do you guys have another idea/approach how to enable scrollbars in snapshots?

PR #877, opened 2025-10-16
feat: add base chat messages

Killusions:
Adds base chat-messages (holds content, actions and attachments), in preparation for next step (chat-container and input)

Consists of one generic chat-message (declarative) and two factories (ai and user message), as well as the attachment list.

The user-message and ai-message components are the building blocks for the chat-container.

Uses #915 (merged)

---

- [x] I confirm that this MR follows the [contribution guidelines](https://github.com/siemens/element/blob/main/CONTRIBUTING.md).

PR #600, opened 2025-08-19
feat: add chat container

Killusions:
This adds the chat-container (fully declarative with auto-scroll), potential follow-ups are an ai-chat-container wrapper (model based with a nice signal-based API for streaming / loading state), dogfooded) in #998  (tool message is here #967 )

The chat input is in #948 (merged), the base components (merged) in #877, the file upload directive (merged) in #932

**Rough Architecture** (ignore the ai chat container for now and the tool message, not a part of this)
ai-chat-container

---

- [x] I confirm that this MR follows the [contribution guidelines](https://github.com/siemens/element/blob/main/CONTRIBUTING.md).

PR #948, opened 2025-11-01
feat: add chat-input

Killusions:
Add the chat-input component, needed for #600, uses #877 (merged)

Automatically expands based on text input (as per UX specs).

The actions and the disclaimer can also be provided in a declarative way.

---


Killusions:
@kfenner Should be ready now (but merge after the base chat messages because of the attachment list), the actions and the disclaimer can also be provided in a declarative way, automatically expands based on text input (as per UX specs), the default legal AI disclaimer will be defaulted in the (ai) chat container, as this input is not AI specific

Killusions:
@spike-rabbit @timowolf Anyone of you time to give it a review here? The actions and the disclaimer can be provided both in a model driven (should be used in most cases to follow UX guidelines) and in a declarative way (extensibility for custom actions like model selectors and custom disclaimer displaying)

PR #949, opened 2025-11-03
chore(demo): improve action card examples

chintankavathia:
added separate example for selectable and actionable cards. Also moved disabled card tile so that it is not the first in the example.


src/app/examples/si-card/si-action-card-selectable.ts
```diff
@@ -0,0 +1,29 @@
+/**
+ * Copyright (c) Siemens 2016 - 2025
+ * SPDX-License-Identifier: MIT
+ */
+import { Component, inject } from '@angular/core';
+import { SiActionCardComponent } from '@siemens/element-ng/card';
+import { SiIconModule } from '@siemens/element-ng/icon';
+import { LOG_EVENT } from '@siemens/live-preview';
+
+@Component({
```
spliffone: Can we use change detection strategy OnPush?

PR #977, opened 2025-11-06
chore(si-photo-uploader): improve UX writing on examples

Danisand:
We realized today, that these examples are copied to products and we want to have a good source.

Why are these texts not part of the language pack?

---

### DIFF
--- a/src/app/examples/si-photo-upload/si-photo-upload.html
+++ b/src/app/examples/si-photo-upload/si-photo-upload.html
@@ -5,11 +5,11 @@
         <si-photo-upload
           #photoUpload
           autoColor="true"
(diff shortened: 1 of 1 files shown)


Danisand:
> @Danisand Thanks, looks good to me 👍
> 
> Regarding the language pack question: These are just texts in the demo examples, these are not texts part of the component itself. As explained in the other issue, the language packs only contain texts from our components as we do not (yet) have a scalable approach to add other texts. Since we don't have a good solution right now, I closed that other issue for now.

Hi @kfenner 

I was mentioning that mainly because the `modalDescription`. This text should be already in the component instead of injecting via API as it works for 95% of the use cases. Or is there already a string in the language pack which operates as smart default if `modalDescription` is not provided?

PR #978, opened 2025-11-07
refactor(tree-view): replace h5 with si-h5 class styling

chintankavathia:
Using semantic tag makes no sense in this context.

- Replace h5 element with div for tree item labels
- Apply si-h5 class conditionally only when selection is enabled and item is selected
- Remove h5-specific font size and weight variables from tree-view-variables.scss
- Remove h5 styling rules from si-tree-view-item.component.scss
- Simplify CSS by removing redundant nested selector for selected item heading weight


spike-rabbit (approved):
👍

PR #987, opened 2025-11-08
docs(stackblitz): provide stackblitz button in live-previewer

spliffone:
This feature provide an `Open in Stackbltz` button in the live-previewer. 
The stackblitz example is created dynamically from scratch and only uses the basic node template.

image

**Included:**

- [x] Provide a `Stackblitz` button in live previewer to show the current example (the typescript file should have the selector 'app-sample').
- [x] Provide only the default element theming
- [x] Automatically add all relative referenced files to the example
- [x] The project template is maintained in the live-previewer asset folder (at the moment we only support Angular) 

**Limitations:**

- No brand theming available
- The template need to contain all possible dependencies since the examples do not provide meta data to to add the dependencies specifically for the example
- The same goes for the Angular providers, it is necessary that the app config has all providers included which are required. 

---

### DIFF
--- a/projects/live-preview/assets/angular/src/styles.scss
+++ b/projects/live-preview/assets/angular/src/styles.scss
@@ -1,0 +1,16 @@
+// Use Element Theme
+@use '@siemens/element-theme/src/theme';
+
+// Use Element components
+@use '@siemens/element-ng/element-ng';
+
+@use '@siemens/element-icons/dist/style/siemens-element-icons';
+
+@use 'flag-icons/sass/flag-icons' with (
+  $flag-icons-use-square: false
+);
+
+app-sample {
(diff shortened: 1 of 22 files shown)


projects/live-preview/assets/angular/src/styles.scss
```diff
@@ -0,0 +1,15 @@
+// Use Element Theme
+@use '@siemens/element-theme/src/theme' with (
+  $element-theme-default: 'siemens-brand',
+  $element-themes: (
+    'siemens-brand',
+    'element'
+  )
+);
```
spliffone:
```suggestion
@use '@siemens/element-theme/src/theme';
```

PR #1010, opened 2025-11-12
refactor(dashboards-ng): use actual type instead of any

chintankavathia:
replace any with actual type for better type safety.

### DIFF
--- a/projects/dashboards-ng/module-federation/index.ts
+++ b/projects/dashboards-ng/module-federation/index.ts
@@ -3,7 +3,7 @@
  * SPDX-License-Identifier: MIT
  */
 import { loadRemoteModule } from '@angular-architects/module-federation';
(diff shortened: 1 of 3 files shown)


projects/dashboards-ng/module-federation/index.ts
```diff
@@ -16,8 +16,8 @@ const setupRemoteComponent = <T>(
 ): Observable<ComponentRef<T>> => {
   const result = new Subject<ComponentRef<T>>();
 
-  loadRemoteModule(factory).then(
-    (module: any) => {
+  loadRemoteModule<Type<T>[]>(factory).then(
```
timowolf: I am unsure, why is this a Type. A type is a component, right? This is the module. Or is a module also a Type.

PR #1015, opened 2025-11-12
docs: mark protected fields internal

spliffone:

---

### DIFF
--- a/api-goldens/element-ng/about/index.api.md
+++ b/api-goldens/element-ng/about/index.api.md
@@ -37,24 +37,14 @@ export class SiAboutComponent implements OnInit {
     readonly copyrightDetails: _angular_core.InputSignal<CopyrightDetails | undefined>;
     readonly icon: _angular_core.InputSignal<string | undefined>;
     readonly iconName: _angular_core.InputSignal<string | undefined>;
-    // (undocumented)
-    protected readonly icons: Record<"elementDocument", string>;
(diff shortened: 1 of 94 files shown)


dr-itz:
Completely agree with @spliffone. adding `@internal` to `protected` is the wrong approach. And it only kills DX. We do protected exactly because it's not part of the public API.

IOW protected _must not_ be part of the goldens.

dr-itz:
> Please also explain how `@internal` kills DX.

It adds unnecessary comment lines stretching the code, making it less easy to read. Waste of vertical space.
And also every commit that changes formatting (hello prettier) or adding stuff like `@internal` only adds stuff to the history of making it harder to trace things, i.e. when hunting bugs/figuring out why something is the way it is.

And it really should be the other way around. Everything that is _added_ has to _add value_.

For me, in the current state, the API goldens do _not_ add any value! It's just a step more. As long as it contains _any_ Angular stuff (also the ng internal functions like `ɵcmp`, `ɵfac`, `ɵprov`). I.e. during review if I have to ignore most of the changes to the goldens and look at the public API changes vs. just look at the public API changes in the code, the added value is exactly _zero_. It only costs me more time.

@siemens/siemens-element-members

spike-rabbit:
I found a way to exclude protected easily in the meantime.

The goldens are anyway more about reviewer experience and maybe used to widen the code owners even more.
So this PR is no longer needed.

PR #1016, opened 2025-11-12
fix(datepicker): add translation key for today button default text

dr-itz:

---


dr-itz:
> @dr-itz Why can't you just set the default of the `todayLabel` input signal to `t(() => $localize`:@@SI_DATEPICKER.TODAY:Today`)` ?

because it's set from the outer component from the datepicker-config so it would be undefined

PR #1032, opened 2025-11-17
fix(filtered-search): avoid triggering  dosearch event incorrectly

akashsonune:

---


spike-rabbit:
This is connected to a longer discussion. I would also be in favor of dropping this behavior. But some people of objections here, as the gitlab search behaves different.

akashsonune:
> This is connected to a longer discussion. I would also be in favor of dropping this behavior. But some people of objections here, as the gitlab search behaves different.

@spike-rabbit Ok then I will close the PR, anyway we have the issue up for the discussion

PR #1034, opened 2025-11-17
fix(application-header): truncate heading if needed

spike-rabbit:
Truncates the application-header heading instead of collapsing rows.
This only applies to the mobile mode. Apps are responsible to switch to it, if there is not enough space in desktop mode.
Adjust the e2e tests to check the truncation.

---

## Summary by CodeRabbit

* **Style**
  * Improved application header layout with text truncation for the logo and application name
  * Adjusted header spacing and overflow behavior to prevent layout breaks on narrow and wide screens

* **Tests**
  * Updated small-screen viewport used in end-to-end tests to better reflect mobile dimensions


timowolf:
> As written in the PR description, this is intended to only work for mobile sizes.
> 
> It is impossible to achieve this with a CSS only solution for desktops. According to the design, we must center the navigation menu in the screen. So we use flex: `1 1 33.33%` for this. But with that, text is truncated before the flex-element grows beyond the 33.33%.

Ok, as it is better than before let us take it. Many thanks.

timowolf (approved):

PR #1035, opened 2025-11-17
fix(datepicker): fix incorrect week numbers

dr-itz:

---

## Summary by CodeRabbit

* **New Features**
  * Datepicker exposes week-of-year lookup (ISO week numbers) for calendar selections.

* **Improvements**
  * More accurate ISO week calculation across year boundaries and edge dates, with configurable week-start handling.

* **Tests**
  * Added tests covering multiple year/week scenarios to validate week-number accuracy.

### DIFF
--- a/projects/element-ng/datepicker/date-time-helper.ts
+++ b/projects/element-ng/datepicker/date-time-helper.ts
@@ -107,8 +107,9 @@ export const getFirstDayInMonth = (year: number, month: number): Date =>
   new Date(year, month - 1, 1);
 
 /**
- * Gets the week number of the specified date.
- * Week number according to the ISO-8601 standard, weeks starting on Monday.
+ * Returns the ISO-8601 week number for the user's week containing the specified date.
+ * ISO-8601 weeks always start on Monday. When the user's week starts on Sunday or Saturday,
+ * the date is adjusted to find the corresponding ISO week (based on the Thursday rule).
  * The first week of a year is the week that contains the first Thursday of the year (='First 4-day week').
  * The highest week number in a year is either 52 or 53.
  *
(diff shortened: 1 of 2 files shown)


akashsonune:
> @akashsonune please review and test carefully. Anything date/time related is always fragile

@dr-itz how do I reproduce the issue, in the demo, I see the week numbers are correct

akashsonune (commented):
@dr-itz I think this should be the week 1 instead of 52?
Image

akashsonune (approved):

projects/element-ng/datepicker/date-time-helper.ts
```diff
@@ -118,16 +118,23 @@ export const getFirstDayInMonth = (year: number, month: number): Date =>
  */
 export const getWeekOfYear = (date: Date, weekStart: WeekStart): number => {
   // Algorithm rewritten from C# example given at http://en.wikipedia.org/wiki/Talk:ISO_week_date
```
akashsonune: Ok instead, maybe a comment here
```suggestion
  // Algorithm rewritten from C# example given at http://en.wikipedia.org/wiki/Talk:ISO_week_date
  // The algorithm is also adjusted for Sunday/Saturday starts
```

dr-itz:
> > > Seems to work fine.
> > > Wouldn't it be breaking for the consumer who might have relied their logic on week number?
> > 
> > 
> > I don't see what could possibly break. The week numbers were calculated wrongly before, now it's correct, so...
> 
> Yeah, but maybe we should add a NOTE or something so that users are aware that the week numbers were wrong earlier, and may adjust their logic in case they rely on it

it will be in the changelog as "fix", so a hard NO

PR #1095, opened 2025-11-27
chore(deps): bump angular dependency

chintankavathia:
See https://github.com/angular/angular/security/advisories/GHSA-58c5-g7wp-6w37


spike-rabbit:
I don't think we need to put an effort here, as this does not effect us. We are not using the XSRF protection by angular in any demo. https://angular.dev/best-practices/security#httpclient-xsrf-csrf-security

chintankavathia:
Closing it as its no longer needed

PR #1105, opened 2025-12-01
docs(tooltips): update ux guidelines

panch1739:
This PR updates the tooltip docs based on the latest Element decisions.

@kfenner This should only be merged when we implement the tooltips coherently in our components. Im unsure if we plan to do it for V49...change it if is not the case (L)

---

- [ x] I confirm that this MR follows the [contribution guidelines](https://github.com/siemens/element/blob/main/CONTRIBUTING.md).


dauriamarco (commented):
Thanks @panch1739 for the update and for the research you did on the topic 🚀

Replying to @kfenner as well:
I’m currently updating my old tooltip PR #778, also as a follow-up to the discussion with Maxi [here](https://github.com/siemens/element/pull/778#issuecomment-3355380529) which should be directly related to this topic.

That PR will very likely be breaking, since the input triggers will be completely removed in order to make the tooltip compliant with the new guidelines.

I’ll keep you posted once the update is ready 👍

---

UPDATE: I've opened a new pull request to address the alignment: #1245

docs/components/status-notifications/tooltip.md
```diff
@@ -19,12 +28,14 @@ Since tooltips are temporary and generally not accessible on touch devices,
...
-- Use sparingly. Avoid overloading the interface with too many tooltips.
-- Use tooltips only on interactive, **keyboard-focusable** elements to ensure accessibility.
+- Do not include interactive elements.
+- Use sparingly to reduce visual noise.
+- Apply tooltips only to interactive, keyboard-focusable elements.
+- Never add tooltips in disabled elements.
 - When implementing tooltips on touch devices, ensure they appear on tap and that this interaction
   does not conflict with other functionalities.
+- Always use the tooltip component, not the browser’s native tooltip.
+- Use a 500 ms show delay and dismiss on pointer leave or focus loss.
```
spike-rabbit:
```suggestion
- Use a 500ms show delay and dismiss on pointer leave or focus loss.
```

Maybe I saw too much code in my life, but I would write it that way

PR #1123, opened 2025-12-02
fix(summary-chip): make value attribute optional

ljanner:
The value attribute is intended to be optional by design, allowing the chip to be used with just a label when no value is needed.

---

## Summary by CodeRabbit

* **New Features**
  * Summary chip value is now optional — chips can render with label and/or icon only.

* **Bug Fixes**
  * Label visibility refined so labels hide/show correctly when both icon and value presence change.

* **Documentation**
  * Clarified guidance: labels are hidden only when both icon and value are present.

* **Tests**
  * Updated snapshots to use explicit checkbox labels and adjusted component tests for label-only chips.

* **Examples**
  * Example pages updated to include chips without values and additional label-only chips.


projects/element-ng/summary-chip/si-summary-chip.component.html
```diff
@@ -18,8 +18,11 @@
       }
     </span>
   }
-  <div class="si-body" [class.visually-hidden]="hideLabel()">
+  @let value = this.value();
+  <div class="si-body" [class.visually-hidden]="hideLabel() && value && actualIcon.icon">
```
ljanner: The conditions with the `&& value && actualIcon.icon` is not really required but makes the component safer to accidental misuse, because purely from API a summary-chip with just a label and the `hideLabel` could be constructed which does not make sense. 

This also aligns with the variants: https://element.siemens.io/components/status-notifications/summary-chip/#variants

spike-rabbit (approved):
👍

PR #1139, opened 2025-12-03
refactor(modal): remove deprecations

akashsonune:
BREAKING CHANGE: Removed property `ModalOptions.initialState`. Use `ModalOptions.inputValues` instead.

---

## Removed ModalOptions.initialState deprecation

Removed the deprecated `initialState` property from `ModalOptions` and migrated all usages to `inputValues`.

Changes:
- Deleted `initialState?: Partial<T>` and its deprecation comment from `ModalOptions`.
- Removed `Object.assign`-based application of `initialState` to component instances.
- Removed spreading of `initialState` for TemplateRef modal content; only `inputValues` is used.
- Updated tests and example to use `inputValues` instead of `initialState`.
- Removed test component member `normalProp` and the test scenario that exercised `initialState`.

Notes:
- BREAKING CHANGE: Code using `ModalOptions.initialState` must migrate to `ModalOptions.inputValues`. Unlike `initialState` (which used Object.assign and could set arbitrary properties), `inputValues` only binds to component inputs — behavior differs and may require changes in callers.
- PR description still requests more detail and the contribution checklist line remains unchecked.

### DIFF
--- a/projects/element-ng/modal/si-modal.service.ts
+++ b/projects/element-ng/modal/si-modal.service.ts
@@ -34,12 +34,6 @@ export interface ModalDependencyInjectionOptions {
 
 export interface ModalOptions<T = Record<string, any>> extends ModalDependencyInjectionOptions {
   /**
-   * Assign all values to the target component using `Object.assign(component, initialState)`.
-   *
-   * @deprecated Use {@link inputValues} instead.
-   */
-  initialState?: Partial<T>;
-  /**
    * Use this to assign values to either `@Input()` or `input()` fields of the provided component.
    * If a template is used, the values are available in the template context.
    */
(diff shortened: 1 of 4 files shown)


projects/element-ng/modal/si-modal.service.ts
```diff
@@ -159,7 +152,7 @@ export class SiModalService {
       environmentInjector: modalRef.data.environmentInjector ?? this.environmentInjector,
       elementInjector: injector
     });
-    Object.assign(modalRef.componentRef.instance as any, modalRef.data?.initialState);
+    Object.assign(modalRef.componentRef.instance as any, modalRef.data?.inputValues);
```
dr-itz: and this one needs to go as we use `setInput` on those below. A component's input should not be assignable when using signal inputs, i.e. should be marked `readonly`. When missed, this could possibly cause unwanted side-effects.

TL;DR: drop this line


```suggestion
```

dr-itz (changes requested):

projects/element-ng/modal/si-modal.service.ts
akashsonune: 👍

dr-itz (approved):

akashsonune:
> @akashsonune please be a little more descriptive in the commit message. Maybe the deprecation has some more details. But beside the renaming, the main difference is, that inputValues is only binding to inputs.

Agree, I updated the BR change note. can you check if it is fine

PR #1167, opened 2025-12-09
refactor(dashboards-ng): remove `initCompleted` output event

chintankavathia:
`initCompleted` has no functional use but it was used just for handling tests. adjusted tests with immediate promise resolver so we don't have to rely on explicit `initCompleted` event subscriptions.

## Refactor: Remove `initCompleted` output event

Removes the unused `initCompleted` output event from `SiWidgetHostComponent` that served no functional purpose and was only used for testing. Tests have been refactored to eliminate their dependency on this event by using Jasmine clock ticks, `fixture.whenStable()`, and Promise-based control flow instead of event subscriptions.

**Changes:**
- Removed `initCompleted` output declaration and its two emission sites from `SiWidgetHostComponent`
- Refactored widget host tests to use Jasmine clock installation/ticking, `fixture.whenStable()`, and explicit change detection instead of subscribing to initialization events
- Refactored grid tests to use async/await patterns with `fixture.whenStable()` for better control flow
- Updated test widget module loader to use `Promise.resolve()` for synchronous immediate resolution instead of dynamic imports

### DIFF
--- a/projects/dashboards-ng/src/components/grid/si-grid.component.spec.ts
+++ b/projects/dashboards-ng/src/components/grid/si-grid.component.spec.ts
@@ -219,7 +219,7 @@ describe('SiGridComponent', () => {
       fixture.componentRef.setInput('widgetCatalog', [TEST_WIDGET]);
       fixture.detectChanges();
       await fixture.whenStable();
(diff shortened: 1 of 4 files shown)


spike-rabbit (approved):
👍

PR #1170, opened 2025-12-09
refactor(split): change default collapse others part behavior

akashsonune:
BREAKING CHANGE: The default value of `SiSplitPartComponent.collapseOthers` input has been changed from `true` to `false`. Previously, when a split part was collapsed, all split parts between it and the end of the split in the respective direction would also collapse automatically. Now, by default, only the individual split part will collapse.

---


spike-rabbit (approved):
👍

PR #1173, opened 2025-12-10
fix(live-preview): support zoneless

chintankavathia:

## Summary

Adds OnPush change detection strategy and explicit change detection triggers across the live-preview module in preparation for zoneless support. The changes include:

- **Import updates**: Adds `ChangeDetectionStrategy` and `ChangeDetectorRef` to multiple components
- **Component metadata**: Marks all live-preview components (si-live-preview, si-live-preview-iframe, si-live-preview-wrapper, si-live-preview-renderer, si-live-preview-qr, si-live-preview-web) and example components with `changeDetection: ChangeDetectionStrategy.OnPush`
- **Change detection triggers**: Injects `ChangeDetectorRef` and adds `markForCheck()` calls after state mutations and message handling in:
  - Route parameter subscriptions
  - Template updates
  - In-progress state changes
  - Fullscreen and collapse toggles
  - Component rendering completion

**Critical Review Feedback**: A major concern was raised about backward compatibility. Marking core live-preview components with OnPush breaks support for existing users and internal libraries that depend on the Default change detection strategy. The reviewer recommends reverting the OnPush decorator from live-preview components themselves while retaining the `markForCheck()` calls for zoneless compatibility, and splitting this into two separate PRs: one for the change detection triggers (with OnPush only on example components) and another specifically for zoneless support.

### DIFF
--- a/projects/live-preview/components/si-live-preview-renderer/si-live-preview-renderer.component.ts
+++ b/projects/live-preview/components/si-live-preview-renderer/si-live-preview-renderer.component.ts
@@ -120,6 +120,7 @@ export class SiLivePreviewRendererComponent implements OnChanges, OnDestroy {
   private internalConfig = inject(SI_LIVE_PREVIEW_INTERNALS);
   private activatedRoute = inject(ActivatedRoute);
   private defaultRoutes = this.activatedRoute.routeConfig?.children ?? [];
+  private cdRef = inject(ChangeDetectorRef);
 
   ngOnChanges(changes: SimpleChanges): void {
     if (changes.exampleUrl?.currentValue) {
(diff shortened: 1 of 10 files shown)


spike-rabbit (changes requested):

dr-itz:
one major problem here: when marking live-previewer components OnPush, any example that is not OnPush will break, i.e. we will lose support for Default CD

projects/live-preview/components/si-live-preview-renderer/si-live-preview-renderer.component.ts
```diff
@@ -301,10 +312,6 @@ export class SiLivePreviewRendererComponent implements OnChanges, OnDestroy {
       private container!: ViewContainerRef;
       private childRouteBackup: Routes | undefined;
 
-      constructor() {
-        this.changeDetector.detach();
```
chintankavathia: So this line was the reason why change detection was not working correctly when switching to zoneless or all onpush. I am not sure why it's added here @dr-itz May be you know. Also I tested some examples keeping on default change detection and that also seems to work but @dr-itz you might have better idea when they fail.

spike-rabbit (approved):
👍

PR #1178, opened 2025-12-10
test(filtered-search): remove irrelevant test

akashsonune:

---

Removes a test case titled "should emit criterion with value = label from config" from the filtered-search component spec file. The removed test verified that when selecting a country option with a value-label pair (e.g., `{ label: 'Switzerland' }` without an explicit `value` property), the emitted criterion would use the label as the value. The test was marked as non-sense in TODO comments, noting that the interface does not allow empty values for options and was scheduled for removal in version 47.

### DIFF
--- a/projects/element-ng/filtered-search/si-filtered-search.component.spec.ts
+++ b/projects/element-ng/filtered-search/si-filtered-search.component.spec.ts
@@ -1997,37 +1997,6 @@ describe('SiFilteredSearchComponent', () => {
       });
     });
 
-    // TODO: this is non-sense. Our interface does not allow empty value for options.
-    // TODO: remove with v47
-    it('should emit criterion with value = label from config', async () => {
-      component.doSearchOnInputChange = true;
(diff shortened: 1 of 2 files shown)


akashsonune:
> @akashsonune somewhere in the actual code, there must also be a check for empty options. I guess you can remove this now. But I don't know where exactly.

@spike-rabbit I guess you mean these changes that I just pushed?

spike-rabbit:
I think just the `?? optionValue.label` part. I thing the rest must be kept.

akashsonune:
> I think just the `?? optionValue.label` part. I thing the rest must be kept.

In that case , we dont see the value field populated in the output event when the value is empty, as the above code ultimately assigns `undefined` to `this.criterionValue().value` (user can still set empty value through template)

spike-rabbit (approved):
👍 

We already have an issue for those kind of empty values. The plan is support them in the future.

PR #1180, opened 2025-12-10
refactor(select): remove deprecated output

akashsonune:
BREAKING CHANGE: Removed deprecated output `SiSelectComponent.dropdownClose`. Use output `SiSelectComponent.openChange` instead.

---

This PR removes deprecated select APIs and tightens option types.

Breaking changes
- Removed SiSelectComponent.dropdownClose output — use SiSelectComponent.openChange instead.
- Removed SelectOptionLegacy interface — use SelectOption / SelectItem (option shape: { type: 'option', value, label, icon?, iconColor?, disabled? }).

Key changes
- Deleted SelectOptionLegacy from the public API and removed all imports/references.
- Removed dropdownClose output and its emit call from SiSelectComponent and updated component metadata.
- Simplified SiSelectSimpleOptionsDirective to accept only SelectItem<T>[] and removed legacy mapping logic.
- Narrowed option types across threshold and readonly-threshold components to SelectOption<T>[] and simplified label/color resolution.
- Updated tests, examples, and snapshots to the new SelectOption/SelectItem shape (value/label/icon/iconColor).

Impact
- Consumers must migrate any dropdownClose listeners to openChange and update legacy option data/types to the new SelectOption/SelectItem shapes before upgrading.


projects/element-ng/select/options/si-select-simple-options.directive.ts
```diff
@@ -42,18 +42,7 @@ export class SiSelectSimpleOptionsDirective<T = string> extends SiSelectOptionsS
   override readonly allRows = computed(() => {
```
spike-rabbit:
```suggestion
  override readonly allRows = computed(() => this.options() ?? [])
```

PR #1187, opened 2025-12-12
fix(icon): render font icons when name contains kebab-case with numbers

spliffone:

---

Fixes icon rendering for kebab-case names containing numbers. Adds a helper method `isKebabCase()` to detect if a string is already in kebab-case format, and modifies `camelToKebabCase()` to return the input unchanged when it's already kebab-case, avoiding unnecessary string transformations. Includes a test case verifying that kebab-case icon names like `element-co2-outdoor` are correctly rendered as icon-font elements.


spike-rabbit (approved):
👍

PR #1203, opened 2025-12-16
feat(schematics): import removal

akashsonune:

---

## Overview
Adds import-removal capability to the schematics migration system to remove specific symbols from import statements during migrations.

## Key Changes
- New data: `ImportRemovalInstruction` and `IMPORT_REMOVALS_MIGRATION` in `projects/element-ng/schematics/migrations/data/import-removals.ts` (removes `buildTrackByIdentity` and `buildTrackByIndex` from @siemens|@simpl element-ng imports).
- Migration integration: `element-migration.ts` processes `importRemovalChanges` from migration data, computes and applies import edits, and inserts adjusted import declarations when needed.
- Utility: `removeImportSymbols()` and `ImportRemovalChange` added to `projects/element-ng/schematics/utils/ts-utils.ts` to compute removal edits for full or partial import deletions.
- API/data wiring: `IMPORT_REMOVALS_MIGRATION` exposed via `migrations/data/index.ts` and included in `getElementMigrationData()`.
- Tests & templates: added test case and template files (`import-removal-inline.template.ts`, expected output) validating removal of the two symbols.

## Impact
Allows declarative removal of deprecated/renamed import symbols during migrations, handling both complete and partial import statements.

### DIFF
--- a/projects/element-ng/schematics/migrations/data/import-removals.ts
+++ b/projects/element-ng/schematics/migrations/data/import-removals.ts
@@ -1,0 +1,21 @@
+/**
+ * Copyright (c) Siemens 2016 - 2025
+ * SPDX-License-Identifier: MIT
+ */
+
+export interface ImportRemovalInstruction {
+  /** Module that the symbol was removed from. */
+  module: RegExp;
+  /** Names of the symbols to be removed from imports. */
+  symbolNames: string[];
+}
+
+/**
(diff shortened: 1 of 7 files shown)


spike-rabbit:
I am not convinced that we add value by just removing the imports. If products have the import they most likely also use it.

So we either need to replace the usage as well or we don't do anything at all.

I am in favor of not doing anything.

PR #1208, opened 2025-12-16
feat(live-preview): add AOT mode

spike-rabbit:
**Warning**: This is currently missing a toggle. So it is always compiled as AOT.

Adds an AOT mode to the live-preview.
This no longer allows editing.

---

### DIFF
--- a/angular.json
+++ b/angular.json
@@ -59,7 +59,8 @@
             "namedChunks": true,
             "define": {
               "maptilerKey": "''",
-              "maptilerUrl": "''"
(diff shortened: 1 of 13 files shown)


chintankavathia (commented):
We would always need aot rendering on our docs and deployed playground. so for what purpose do we want to still keep JIT variant? also the JIT is not working well with zoneless so it would create even more confusion. I would just drop jit even if that means no edit mode.

spike-rabbit:
we have AOT

PR #1238, opened 2025-12-22
fix(select): maintain filtered rows when updating all rows

spike-rabbit:
This commit turns the rows 
(which represents the rows that are visible)
for eagerly loaded options into a computed signal.
In addition, it stores the current filter value,
so we no longer need to call the filter function
when new options are provided.
Thus, the filter value is maintained and the rows
are anyway updated based on the new options.

Co-authored-by: Bajohr, Rayk <rayk.bajohr@siemens.com>

---

## Summary

This PR fixes the select component to preserve the active filter and maintain filtered (visible) rows when all options are updated.

### Key changes

- Removed the `OnChanges` lifecycle hook from `SiSelectComplexOptionsDirective` and `SiSelectSimpleOptionsDirective`.
- Refactored filtering in `SiSelectOptionsStrategyBase`:
  - Added an abstract `allRows` to represent all options.
  - Converted `rows` from a writable signal to a computed signal derived from `allRows` and an internal `filterValue`.
  - `onFilter` now updates the internal `filterValue` instead of mutating `rows`.
- Added a test verifying the current filter is applied when options are updated.

### Impact

Visible rows now preserve the active filter when new options are provided, avoiding unnecessary re-filtering or reset and ensuring consistent filter state across updates.

### DIFF
--- a/api-goldens/element-ng/select/index.api.md
+++ b/api-goldens/element-ng/select/index.api.md
@@ -83,14 +83,12 @@ export class SiSelectActionsDirective {
 }
 
 // @public @deprecated
-export class SiSelectComplexOptionsDirective<T> extends SiSelectOptionsStrategyBase<T> implements OnChanges {
+export class SiSelectComplexOptionsDirective<T> extends SiSelectOptionsStrategyBase<T> {
     // (undocumented)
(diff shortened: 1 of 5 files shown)


kfenner (approved):
LGTM 👍

PR #1250, opened 2026-01-05
fix(tree-view): use hasAnyChildren from parent to determine inline padding

chintankavathia:
caching `hasAnyChildren` in variable during initialization doesn't always sync with parent and causes unexpected padding calculation when tree items are filtered and added again.

Removed the cached displayFolderState and replaced it with direct checks of this.treeViewComponent.hasAnyChildren. The getters showFolderStateStart and showFolderStateEnd now read the parent's hasAnyChildren dynamically so inline padding reflects the current parent state (fixes incorrect padding when items are filtered and re-added).


spike-rabbit (approved):
👍

PR #1252, opened 2026-01-05
fix(select): maintain filtered rows when updating all rows

spike-rabbit:
This commit turns the rows
(which represents the rows that are visible)
for eagerly loaded options into a computed signal. In addition, it stores the current filter value,
so we no longer need to call the filter function
when new options are provided.
Thus, the filter value is maintained and the rows
are anyway updated based on the new options.

(backport from https://github.com/siemens/element/pull/1238)
(cherry picked from commit 769a8f72bb0a5aefafbd9e1a94cd8c0f3bf54e3c)

---

### DIFF
--- a/api-goldens/element-ng/select/index.api.md
+++ b/api-goldens/element-ng/select/index.api.md
@@ -85,14 +85,12 @@ export class SiSelectActionsDirective {
 }
 
 // @public @deprecated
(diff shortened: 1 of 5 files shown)


kfenner (approved):
LGTM 👍

PR #1261, opened 2026-01-07
refactor(date-range-filter): use signal `model` for value input

spike-rabbit:
Use a `model` signal for the `value` input.
Remove the redundant `internalValue` signal.

This is a linter finding in a newer version.

---

Refactors SiRelativeDateComponent to use a signal-based model for value handling and updates tests.

Changes:
- Replaces the public `value` input with `readonly value = model(0)` (signal).
- Removes `valueChange` output and the `internalValue` signal.
- Simplifies ngOnChanges to trigger calculateOffset on value changes.
- calculateOffset, updateValue, and changeUnit now read/write `value()` directly (no output emissions).
- Tests updated to use two-way binding [(value)]="value" with a host signal.

Motivation: addresses linter findings for signal-based patterns; contributor confirms compliance with contribution guidelines.

Known issue: reviewers reported a visual-regression where displayed range doesn't match the selected "after" value when switching modes; PR #1262 was referenced but did not resolve it — reviewer expects the issue to be fixed once that follow-up is merged.

### DIFF
--- a/projects/element-ng/date-range-filter/si-relative-date.component.spec.ts
+++ b/projects/element-ng/date-range-filter/si-relative-date.component.spec.ts
@@ -20,10 +20,9 @@ const ONE_DAY = ONE_MINUTE * 60 * 24;
   imports: [SiRelativeDateComponent],
   template: `<si-relative-date
     [enableTimeSelection]="enableTimeSelection"
-    [value]="value()"
     [unitLabel]="unitLabel"
     [valueLabel]="valueLabel"
-    (valueChange)="value.set($event)"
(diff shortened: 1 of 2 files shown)


dr-itz:
> > Well, VRT is correct here, it's broken: the displayed date range doesn't match the selected "after", works on main. But then again, coming from simple mode to advanced mode should actually keep the range, i.e. either use "before" mode or use after mode but with the start date as reference point. Seems to be a recent breakage, this used to work
> 
> #1262 should help

it doesn't..something is completely broken

PR #1263, opened 2026-01-07
docs(buttons): update buttons examples

dauriamarco:
Updates and improves all buttons examples including `buttoncategories` according to the new `btn-icon` specs.

---


src/app/examples/buttons/button-groups.html
```diff
@@ -76,6 +76,18 @@
         <i class="icon element-delete"></i>
       </button>
     </div>
+
+    <div class="btn-group" role="group" aria-label="Tertiary group">
+      <button type="button" class="btn btn-icon btn-ghost" aria-label="Edit">
```
dauriamarco: Hi @panch1739, quick question for the button group: in the case where it contains icons only, can it also be presented using the ghost button variant?

image

PR #1338, opened 2026-01-20
fix(side-panel): set default display to inline-flex

dauriamarco:
Sets the default display to `inline-flex` for the fullscreen button in the side panel.

---


spike-rabbit (approved):
👍

PR #1350, opened 2026-01-21
refactor(vertical-navbar): use native CSS animations for si-navbar-ve…

spliffone:
…rtical-header

---

### DIFF
--- a/projects/element-ng/navbar-vertical/si-navbar-vertical-header.component.scss
+++ b/projects/element-ng/navbar-vertical/si-navbar-vertical-header.component.scss
@@ -1,12 +1,57 @@
 @use '@siemens/element-theme/src/styles/variables';
 @use 'sass:map';
 
-:host {
-  display: flex;
-  align-items: center;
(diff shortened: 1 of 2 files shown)


projects/element-ng/navbar-vertical/si-navbar-vertical-header.component.scss
```diff
@@ -1,12 +1,45 @@
...
+    opacity: 0;
+    visibility: hidden;
+    transition: opacity $content-duration ease-in visibility $content-duration;
+  }
+
+  .header-content {
+    opacity: 1;
+    visibility: visible;
+    transition: opacity $content-duration ease-out visibility $content-duration;
+    transition-delay: 0.3s;
```
spike-rabbit: I think you can also include this in the shorthand notation

spike-rabbit (changes requested):
Did you check whether it might be easier to just use animate.enter / leave instead of always having those elements in the DOM?

PR #1365, opened 2026-01-23
refactor(form): make errrormessageid required

akashsonune:
BREAKING CHANGE: The property `SiFormItemControl.errormessageId` is now required. All implementations of `SiFormItemControl` must provide this property to ensure proper accessibility support for error messages.

---


projects/element-ng/form/si-form-item/si-form-field-native.control.ts
```diff
@@ -46,6 +46,8 @@ export class SiFormFieldNativeControl implements SiFormItemControl {
 
   isFormCheck: boolean;
 
+  readonly labelledby = null;
```
spike-rabbit: I guess this change is no longer needed

spike-rabbit (changes requested):
Please also add a short explanation to the label commit, that `si-form-items` without a label are not a11y conform. So we made it required.

spike-rabbit (approved):
👍

Issue #1373, opened 2026-01-26
forms: provide the the min/max value in validator error messages


michael-smt:
### Prerequisites

- [x] I have read the [Contributing Guidelines](https://github.com/siemens/element/blob/main/CONTRIBUTING.md).
- [x] I have not leaked any internal/restricted information like screenshots, videos, code snippets, links etc.

### Feature request

For the the `min*` and `max*` validators, the validator error messages from the `SiFormValidationErrorMapper` should include information about what the smallest or biggest accepted value is.

For example instead of `The value is too small` the message could be `The value should be {{min}} or more`.

Affected error messages and translation keys:
- [ ] `min` / `SI_FORM_CONTAINER.ERROR.MIN`
- [ ] `max` / `SI_FORM_CONTAINER.ERROR.MAX`
- [ ] `minlength` / `SI_FORM_CONTAINER.ERROR.MIN_LENGTH`
- [ ] `maxlength` / `SI_FORM_CONTAINER.ERROR.MAX_LENGTH`
- [ ] `minDate` / `SI_FORM_CONTAINER.ERROR.MIN_DATE`
- [ ] `maxDate` / `SI_FORM_CONTAINER.ERROR.MAX_DATE`
- [ ] `minTime` / `SI_FORM_CONTAINER.ERROR.MIN_TIME`
- [ ] `maxTime` / `SI_FORM_CONTAINER.ERROR.MAX_TIME`

### Use case

The user sees what values are expected.

PR #1374, opened 2026-01-26
feat(form): extend and align validation messages with UX guidelines

spliffone:
Fixing #1373

---


projects/element-ng/form/si-form-validation-error.service.ts
```diff
@@ -16,46 +16,61 @@ import { SiFormValidationErrorMapper } from './si-form-validation-error.model';
...
+    () => $localize`:@@SI_FORM_CONTAINER.ERROR.MAX_LENGTH:Max. {{requiredLength}} characters`
+  ),
+  milliseconds: t(
+    () => $localize`:@@SI_FORM_CONTAINER.ERROR.MILLISECONDS:Integer between 0 and 999 required`
+  ),
+  min: t(() => $localize`:@@SI_FORM_CONTAINER.ERROR.MIN:Min. {{min}}`),
+  minDate: t(
+    () => $localize`:@@SI_FORM_CONTAINER.ERROR.MIN_DATE:Date after {{minString}} required`
+  ),
+  minTime: t(() => $localize`:@@SI_FORM_CONTAINER.ERROR.MIN_TIME:The time is too far in the past.`),
```
Danisand: @spliffone we need to add a variable


```suggestion
  minTime: t(() => $localize`:@@SI_FORM_CONTAINER.ERROR.MIN_TIME:Time after {{minString}} required`),
```

Danisand:
Hi @spliffone and @michael-smt 

I have seen the picture for the visual regression test. It seems that the key `SI_FORM_CONTAINER.ERROR.PATTERN` is concatenated in the tooltip.

Could you please explain me, in which situation `SI_FORM_CONTAINER.ERROR.PATTERN` is used?

spliffone:
> Hi @spliffone and @michael-smt
> 
> I have seen the picture for the visual regression test. It seems that the key `SI_FORM_CONTAINER.ERROR.PATTERN` is concatenated in the tooltip.
> 
> Could you please explain me, in which situation `SI_FORM_CONTAINER.ERROR.PATTERN` is used?

Yes, the error is provided when the app apply the Angular pattern directive (https://angular.dev/api/forms/PatternValidator#adding-a-pattern-validator)

spliffone:
> > @spike-rabbit I think we should replace the current div's inside the `.invalid-feedback` container by lists `ul` / `li`. I guess this way the accessibility tree would show them not as one concatenated string.
> 
> I don't think we should do that: image (see: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-describedby)
> 
> Outside of this PR we could start evaluating:
> 
> * if moving to aria-errormessage works
> * reference each error seperately

The PR has enough changes I will create an issue where we should discuss it first.

PR #1421, opened 2026-02-03
feat(element-ng): replace `@angular/animations` with native CSS

spike-rabbit:
All usages of `@angular/animations` (deprecated) were removed.
With this commit we remove the entire package.

BREAKING CHANGE: Animations can no longer be disabled using
`@angular/animations` specific features like:
- `NoopAnimationsModule`
- `BrowserAnimationsModule.withConfig({disableAnimations: false})`
- `provideNoopAnimations`
- `provideAnimationsAsync('noop')`

Use CSS to disable animations. See: https://element.siemens.io/architecture/motion-animation/

---


chintankavathia (changes requested):
Just one small thing from my side other wise looks good.

PR #1425, opened 2026-02-03
refactor(tabs): use `siTooltip` instead of `title`

dauriamarco:
Adds tooltip support to tabs.

---


spike-rabbit (changes requested):

src/app/examples/si-tabs/si-tabs-routing.html
```diff
@@ -1,7 +1,7 @@
 <si-tabset>
   <a si-tab heading="Home" [routerLink]="tabLink1"></a>
   <a si-tab heading="Energy" [routerLink]="tabLink2"></a>
-  <a si-tab heading="Test coverage" [routerLink]="tabLink3"></a>
+  <a si-tab heading="Test coverage" icon="element-alchemy-flask" [routerLink]="tabLink3"></a>
```
chintankavathia: Either all tabs should be icon or text. we shouldn't show example with mix of both as per our UX guideline.

chintankavathia (changes requested):

src/app/examples/si-tabs/si-tabs-routing.html
chintankavathia: Yes. I guess they came via migration from legacy examples.

spike-rabbit:
implemented as Part of #2231

PR #1426, opened 2026-02-03
feat(buttons): add tooltip support to selection buttons

dauriamarco:
Adds tooltip support to selection buttons so that they can support tooltips in all configurations both when the label wraps the input/element (text, icon, etc.) and when there’s no wrapping label (using the for attribute).

---


src/app/examples/buttons/selection-buttons.html
```diff
@@ -33,19 +33,44 @@ <h4>Single-select</h4>
 <br /><br />
 
 <div class="btn-group" role="group" aria-label="Layout selection toggle">
-  <input type="radio" class="btn-check" name="btnradio3" id="btnradio3-1" checked />
-  <label class="btn btn-icon" for="btnradio3-1" aria-label="One pane">
-    <i class="icon element-layout-pane-1" aria-hidden="true"></i>
+  <label>
+    <input
```
spike-rabbit:
```suggestion
  <label class="btn btn-icon">
    <input
```

and remove the inner span. Does this work / can you make it work?

spike-rabbit (changes requested):

src/app/examples/buttons/selection-buttons.html
dauriamarco: 👍

PR #1431, opened 2026-02-03
fix(datepicker): update date even on changing only time

chintankavathia:
currently date never updates when only time value is changed in datepicker.

### DIFF
--- a/projects/element-ng/datepicker/si-datepicker.component.ts
+++ b/projects/element-ng/datepicker/si-datepicker.component.ts
@@ -284,10 +284,15 @@ export class SiDatepickerComponent implements OnInit, OnChanges, AfterViewInit {
     nonNullable: true
   });
   /**
+   * Stores the current time value before a new time change is applied.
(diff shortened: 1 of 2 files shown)


projects/element-ng/datepicker/si-datepicker.component.ts
```diff
@@ -511,7 +516,13 @@ export class SiDatepickerComponent implements OnInit, OnChanges, AfterViewInit {
   }
 
   protected timeSelected(newTime?: Date): void {
-    if (newTime == null || newTime === this.time.value) {
+    if (
+      newTime == null ||
+      (newTime.getHours() === this.previousTime?.getHours() &&
```
spliffone: Can we get rid of the comparision and just check for null / undefined?

spliffone: What happens if we adjust the SiTimePickerComponent setTime function to simply verify changes by checking the time `this.time?.getTime() !== time?.getTime()`

PR #1440, opened 2026-02-04
refactor(dashboards): use common util for federation component loading

chintankavathia:
Moves component creation and error handling logic from module-federation, native-federation, mf-bridge into common util.

### DIFF
--- a/projects/dashboards-ng/src/public-api.ts
+++ b/projects/dashboards-ng/src/public-api.ts
@@ -14,6 +14,7 @@ export * from './model/widgets.model';
 export * from './model/si-dashboard-toolbar.model';
 export * from './model/si-widget-id-provider';
 export * from './widget-loader';
+export * from './services/federation-loader.util';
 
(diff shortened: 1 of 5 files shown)


projects/dashboards-ng/src/public-api.ts
```diff
@@ -14,6 +14,7 @@ export * from './model/widgets.model';
 export * from './model/si-dashboard-toolbar.model';
 export * from './model/si-widget-id-provider';
 export * from './widget-loader';
+export * from './services/federation-loader.util';
```
spike-rabbit: Does this really needs to be public? Maybe just add /** @internal */ if only technically needed

spike-rabbit (changes requested):

projects/dashboards-ng/src/public-api.ts
chintankavathia: done

PR #1458, opened 2026-02-05
fix(filtered-search): avoid expression change errors in zoneless apps

chintankavathia:
this doesn't reproduce on our live previewer somehow.
but when running apps in zoneless mode we get the expression change errors.

### DIFF
--- a/projects/element-ng/autocomplete/si-autocomplete.directive.ts
+++ b/projects/element-ng/autocomplete/si-autocomplete.directive.ts
@@ -2,7 +2,7 @@
  * Copyright (c) Siemens 2016 - 2025
  * SPDX-License-Identifier: MIT
  */
-import { Directive, HostBinding, HostListener } from '@angular/core';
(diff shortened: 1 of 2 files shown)


projects/element-ng/autocomplete/si-autocomplete.directive.ts
```diff
@@ -11,35 +11,31 @@ import { SiAutocompleteOptionDirective } from './si-autocomplete-option.directiv
...
   protected get activeDescendant(): string {
-    return this.listbox?.active?.id() ?? '';
+    return this.listbox()?.active?.id() ?? '';
   }
 
-  @HostBinding('attr.aria-controls')
-  protected get ariaControls(): string | undefined {
-    return this.listbox?.id();
-  }
+  protected readonly ariaControls = computed(() => this.listbox()?.id());
```
spike-rabbit: I guess you can also just inline this in the host field. No need for extra computed here

PR #1472, opened 2026-02-06
refactor(collapsible-panel): improve transition for collapsible content

dauriamarco:
Fixes the animation for the collapsible panel content.

---

### DIFF
--- a/projects/element-ng/accordion/si-collapsible-panel.component.scss
+++ b/projects/element-ng/accordion/si-collapsible-panel.component.scss
@@ -99,7 +99,7 @@
 .content-motion {
   display: grid;
   grid-template-rows: 1fr;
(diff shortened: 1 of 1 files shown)


projects/element-ng/accordion/si-collapsible-panel.component.scss
```diff
@@ -108,4 +108,13 @@
...
     grid-template-rows: 0fr;
   }
+
+  :host(.full-height) & {
+    transition-duration: 0s;
+    transition-delay: variables.$element-default-transition-duration;
+
+    @starting-style {
+      grid-template-rows: 1fr;
+    }
```
spike-rabbit: wouldn't it be sufficient if you just do something like


```suggestion
   // Disables the animation on the inner content
    display: flex;
```

spike-rabbit (changes requested):
Please check if this actually works

PR #1524, opened 2026-02-13
refactor(sort-bar): deprecate SiSortBarComponent and SiSortBarModule

chintankavathia:
DEPRECATED: `SiSortBarComponent` and `SiSortBarModule` are deprecated. They originate from the older design system and do not align with current design guidelines. No known use case exists for this component. Both will be removed in v50.


panch1739:
> @spliffone can you share some screen shots how it is used

@spike-rabbit yes, it would be super nice to see...so far, i havent seen anyone using this component 🤷

PR #1580, opened 2026-02-27
test(element): remove TestHostComponent component for launchpad, avatar and card

chintankavathia:

### DIFF
--- a/projects/element-ng/avatar/si-avatar.component.spec.ts
+++ b/projects/element-ng/avatar/si-avatar.component.spec.ts
@@ -2,55 +2,47 @@
  * Copyright (c) Siemens 2016 - 2026
  * SPDX-License-Identifier: MIT
  */
-import { Component } from '@angular/core';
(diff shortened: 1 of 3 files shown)


projects/element-ng/avatar/si-avatar.component.spec.ts
```diff
@@ -2,55 +2,47 @@
...
+  const altText = signal('Test');
+  const status = signal<EntityStatusType | undefined>(undefined);
+  const autoColor = signal(false);
 
   beforeEach(() => {
-    fixture = TestBed.createComponent(TestHostComponent);
-    host = fixture.componentInstance;
-    host.altText = 'Test';
-    element = fixture.nativeElement.querySelector('si-avatar');
+    imageUrl.set(undefined);
```
spliffone: Just an idea, in case we use for the signals a `let` we can directly assign them in the `beforeEach` and we don't need to care about the previous state?

PR #1914, opened 2026-04-21
build(npm): centralize node and npm version

### LINKED ISSUE #1900: Prevent NPM version missmatch on release branches

spliffone:
fixes https://github.com/siemens/element/issues/1900

---

[Documentation](https://d33c9dcnqinn2a.cloudfront.net/pr-1914/pages/).
[Examples](https://d33c9dcnqinn2a.cloudfront.net/pr-1914/pages/element-examples/).
[Dashboards Demo](https://d33c9dcnqinn2a.cloudfront.net/pr-1914/pages/dashboards-demo/).
[Playwright report](https://d33c9dcnqinn2a.cloudfront.net/pr-1914/playwright-report/).

**Coverage Reports:**

Code Coverage

- [charts-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-1914/pages/coverage/charts-ng/)
- [dashboards-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-1914/pages/coverage/dashboards-ng/)
- [element-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-1914/pages/coverage/element-ng/)
- [element-translate-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-1914/pages/coverage/element-translate-ng/)
- [maps-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-1914/pages/coverage/maps-ng/)
- [native-charts-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-1914/pages/coverage/native-charts-ng/)


package.json
```diff
@@ -14,7 +14,7 @@
   },
   "type": "module",
   "engines": {
-    "node": "^24.0.0",
+    "node": "^24.15.0",
     "yarn": "Please use NPM to install dependencies",
     "npm": "11"
```
spike-rabbit, line 19: Should we also fix this? I guess just to the version bundled with the respective node version?

spike-rabbit (changes requested):

package.json
spliffone: I am a bit uncertain we need it since node is bundled with npm already so I would rather vote to drop this entry

spliffone:
We migrated to pnpm so the PR is obsolate

PR #1933, opened 2026-04-22
build(pnpm): migrate to pnpm

spliffone:
See https://github.com/siemens/element/issues/1900

| | PNPM | NPM |
| - | - | - |
| `node_modules` structure |  Non-flat by default — only direct dependencies appear in the root of `node_modules`. Transitive deps are stored under `.pnpm/` and symlinked.  |  Flat/hoisted — all packages (including transitive deps) are hoisted to the root of `node_modules`. | 
| Storage model | Uses a global content-addressable store. Files are hard-linked into node_modules — only one physical copy per version across all projects. Only changed files are added when upgrading a dependency. | Copies every package into each project's node_modules. |
| Isolated `node_modules` | Yes (default) | Yes ([install linked](https://manzoorwani.dev/npms-install-strategylinked-is-getting-better-you-should-try-it/#:~:text=npm's%20install%2Dstrategy%3Dlinked%20takes,own%20node_modules%2F%20to%20the%20store.)) |
| Ships with Node.js | No | Yes |
| Workspace support | Yes | Yes |
| Managing Node.js versions | Yes  (`pnpm env`) | No |
| Catalogs (shared dep versions) | Yes ([catalogs](https://pnpm.io/catalogs)) | No |
| Auto-install before script run | Yes (ensure `node_modules` are in sync with lock file)  | No |
| Supported [JSR registry](https://jsr.io/)  | Yes ([npm-compatibility](https://jsr.io/docs/npm-compatibility)) | No |
| Postinstall allow list | Yes ([Config dependencies](https://pnpm.io/11.x/config-dependencies)) | No |
| Lock file stability | Stable between feature version | Unstable |
| Branch specific log file | Yes ([Git branch lockfiles](https://pnpm.io/git_branch_lockfiles)) | No |
| Installation hooks | Yes ([.pnpmfile.mjs](https://pnpm.io/11.x/pnpmfile)) | No |

- PNPM is consistently faster than NPM (see [PNPM Benchmark results](https://pnpm.io/benchmarks))
- `ng serve` (vite) require a flat `node_modules` to run the dev server which require adjustments in the PNPM configuration to hoist dependencies 
- The generated `tarballUrl` in the lock file are incorrectly for code.siemens.com

**Challenges:**

- Installation of code.siemens packages require the tarBallUrl

---

[Documentation](https://d33c9dcnqinn2a.cloudfront.net/pr-1933/pages/).
[Examples](https://d33c9dcnqinn2a.cloudfront.net/pr-1933/pages/element-examples/).
[Dashboards Demo](https://d33c9dcnqinn2a.cloudfront.net/pr-1933/pages/dashboards-demo/).
[Playwright report](https://d33c9dcnqinn2a.cloudfront.net/pr-1933/playwright-report/).

**Coverage Reports:**

Code Coverage

- [charts-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-1933/pages/coverage/charts-ng/)
- [dashboards-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-1933/pages/coverage/dashboards-ng/)
- [element-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-1933/pages/coverage/element-ng/)
- [element-translate-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-1933/pages/coverage/element-translate-ng/)
- [maps-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-1933/pages/coverage/maps-ng/)
- [native-charts-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-1933/pages/coverage/native-charts-ng/)


package.json
```diff
@@ -180,12 +167,16 @@
...
-      "@siemens/element-theme": "file:projects/element-theme"
+  "pnpm": {
+    "onlyBuiltDependencies": [
+      "@parcel/watcher",
+      "esbuild",
+      "lmdb",
+      "msgpackr-extract"
+    ],
+    "overrides": {
+      "@microsoft/api-extractor>typescript": "5.9.3",
```
spike-rabbit: is there no `$typescript` available?

spike-rabbit (changes requested):
I think we also must use the pnpm semantic-elease plugin. If not npm will install a bunch of things when releasing.
With that:
- you can use alias in the peer dependencies to workspace deps
- I hope it works with the dist packages as well somehow... may needs scripting

chintankavathia:
I remember we discussed about using pnpm in past and one of the reason for not switching was:
So far most of the consumer projects are using yarn and npm. Since npm has the most problems with the dependency management we tend to stay on npm. Since we discover this problem earlier.

What exactly got changed now?

spliffone:
> I remember we discussed about using pnpm in past and one of the reason for not switching was: So far most of the consumer projects are using yarn and npm. Since npm has the most problems with the dependency management we tend to stay on npm. Since we discover this problem earlier.
> 
> What exactly got changed now?

@chintankavathia yes most of our consumer still use `yarn` or `npm` while almost all dependency issues are in `npm` environments. This PR is just an consideration to solve the frequent issue we have with the package-lock.json updates (see also #1914). Just two things:

- with npm we might figure out dependency issues with siemens lint, but since we have a mono-repo workspace I don't believe we find issues related to element.
- I thought the decision to use npm was driven by the semantic-release support 

So let's discuss it in the core-meeting what is the right approach.

package.json
```diff
@@ -164,6 +158,10 @@
     "postcss": "8.5.10",
     "postcss-scss": "4.0.9",
     "prettier": "3.8.3",
+    "prismjs": "1.30.0",
+    "qrcode-generator": "2.0.4",
+    "react": "19.2.5",
+    "react-dom": "19.2.5",
```
spike-rabbit: please confirm here: is this really needed to move live-preview dependencies here?

spike-rabbit (changes requested):
To provide some more details regarding releasing:

In pnpm workspaces one can declare (peer)dependencies using the [workspace: syntax](https://pnpm.io/workspaces). 
With that I hope we can eliminate the use of custom [updating the peer dependencies](https://github.com/siemens/element/blob/main/tools/release/generate-versions.js).

The only problem is, that I don't know how this behaves with releasing from a dist directory, so when the actual workspace package is not the target.

kfenner:
@spliffone @spike-rabbit 3 hours ago, new pnpm major release v11: https://github.com/pnpm/pnpm/releases/tag/v11.0.0

Thanks to @pwuersch

.github/workflows/build-and-test.yaml
```diff
@@ -55,22 +56,23 @@ jobs:
...
           path: dist
-      - run: npm ci --prefer-offline --no-audit
-      - run: npm run translate:test -- --watch=false --progress=false --coverage
-      - run: npm run lib:test -- --watch=false --progress=false --coverage
-      - run: npm run schematics:test
-      - run: npm run charts:test -- --watch=false --progress=false --coverage
-      - run: npm run dashboards:test -- --watch=false --progress=false --coverage
-      - run: npm run maps:test -- --watch=false --progress=false --coverage
-      - run: npm run native-charts:test -- --watch=false --progress=false --coverage
+      - run: pnpm install --frozen-lockfile
```
pwuersch: [`--frozen-lockfile`](https://pnpm.io/cli/install#--frozen-lockfile) is already true for CI environments. GitHub Actions [sets the `CI` variable](https://docs.github.com/en/actions/reference/workflows-and-actions/variables) by default, and pnpm respects that.

[`--prefer-offline`](https://pnpm.io/cli/install#--prefer-offline) can be used to skip some checks, but I am not sure anymore how much this actually saves.

```suggestion
      - run: pnpm install --prefer-offline=true
```

.github/workflows/build-and-test.yaml
```diff
@@ -143,17 +146,18 @@ jobs:
...
+          cache: 'pnpm'
       - uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8
         with:
           name: dist
           path: dist
       # Not injecting the token will exclude the brand packages, but this is fine for e2e tests.
-      - run: npm ci --prefer-offline --no-audit --include=optional
-      - run: npx playwright test --shard=${{ matrix.shardIndex }}/${{ matrix.shardTotal }}
+      - run: pnpm install --frozen-lockfile
+      - run: pnpm exec playwright test --shard=${{ matrix.shardIndex }}/${{ matrix.shardTotal }}
```
pwuersch, line 173: Nit: the `exec` here is technically not required (but can add some clarity I guess): https://pnpm.io/cli/exec#examples

package.json
```diff
@@ -13,10 +13,10 @@
     "email": "info@siemens.com"
   },
   "type": "module",
+  "packageManager": "pnpm@10.6.3",
   "engines": {
     "node": "^24.0.0",
-    "yarn": "Please use NPM to install dependencies",
-    "npm": "11"
+    "pnpm": ">=9"
```
pwuersch: I think pnpm has too many breaking changes to keep this at a `>=9`. The `packageManager` field already restricts you from running a different package manager version when using Corepack. IMO this should be entirely removed here.

```suggestion
```

pwuersch (commented):
Hey guys, chiming in here because @kfenner was asking for some feedback from users that have used pnpm before.

pnpm has sped up our local installs quite a bit since we migrated away from Yarn already three years ago. Probably NPM has become more competitive over the years. The [headlining reasons](https://pnpm.io/motivation) to use pnpm are quite convincing:
- Shared package store, which is faster and more efficient,
- Allow-listing of dependencies that are allowed to execute package lifecycle scripts -> mitigates almost every recent supply-chain attack we've seen, as the compromised package must be in the allowed dependencies to cause damage.

While this is already great, the real benefit in working with projects that use pnpm is the developer experience:
- Unmet `peerDependencies` are printed in an actually useful way,
- It still adheres to most of the vanilla npm configuration (e.g., overriding registry URLs and other settings via `.npmrc`),
- Very nice integration in [`mise-en-place`](https://mise.jdx.dev/mise-cookbook/nodejs.html#example-with-pnpm) or [`corepack`](https://pnpm.io/installation#using-corepack), whichever you prefer,
- Amazing lockfiles with diffs you can actually read.

All in all, I enjoy working with pnpm and it's become my default choice for a package manager when using Node.js.

package.json
spliffone: We have two options:

1. we add them to the root package so they are automatically node_modules top level dependencies
2. we add add them to the specific dependencies to hoist patterns so the are add to node_modules top level dependencies

spliffone:
> @spliffone @spike-rabbit 3 hours ago, new pnpm major release v11: https://github.com/pnpm/pnpm/releases/tag/v11.0.0
> 
> Thanks to @pwuersch

Sadly pnpm 11 handles optionalDependencies differently and I wasn't able to link the brand dependency in the root node_modules folder which means the prepare brand script is always failing and the pipeline never succeed. So I guess we should go with version 10 at the moment.

kfenner:
> > @spliffone @spike-rabbit 3 hours ago, new pnpm major release v11: https://github.com/pnpm/pnpm/releases/tag/v11.0.0
> > Thanks to @pwuersch
> 
> Sadly pnpm 11 handles optionalDependencies differently and I wasn't able to link the brand dependency in the root node_modules folder which means the prepare brand script is always failing and the pipeline never succeed. So I guess we should go with version 10 at the moment.

Hmm, is this a good sign?

pwuersch:
> > @spliffone @spike-rabbit 3 hours ago, new pnpm major release v11: [pnpm/pnpm@`v11.0.0` (release)](https://github.com/pnpm/pnpm/releases/tag/v11.0.0)
> > Thanks to @pwuersch
> 
> Sadly pnpm 11 handles optionalDependencies differently and I wasn't able to link the brand dependency in the root node_modules folder which means the prepare brand script is always failing and the pipeline never succeed. So I guess we should go with version 10 at the moment.

@spliffone locally, with a normal `pnpm install` using pnpm v11.0.0 I have the following file structure:
```text
node_modules/@simpl
├── brand
└── docs-composer
```

Also the `prepare-brand` script succeeds when running `pnpm prepare-brand`. Did you change something in the meantime, or am I missing something?

spliffone:
> > > @spliffone @spike-rabbit 3 hours ago, new pnpm major release v11: [pnpm/pnpm@`v11.0.0` (release)](https://github.com/pnpm/pnpm/releases/tag/v11.0.0)
> > > Thanks to @pwuersch
> > 
> > 
> > Sadly pnpm 11 handles optionalDependencies differently and I wasn't able to link the brand dependency in the root node_modules folder which means the prepare brand script is always failing and the pipeline never succeed. So I guess we should go with version 10 at the moment.
> 
> @spliffone locally, with a normal `pnpm install` using pnpm v11.0.0 I have the following file structure:
> 
> ```
> node_modules/@simpl
> ├── brand
> └── docs-composer
> ```
> 
> Also the `prepare-brand` script succeeds when running `pnpm prepare-brand`. Did you change something in the meantime, or am I missing something?

Yes, I started the migration again and dropped a few settings. Now it seems to work

spliffone:
See https://github.com/siemens/element/issues/1900

**Summary:**

(+) way more powefull than npm
(+) more security
(+) releasing is less complex: no postversion
(+) stable lockfile
(+) no -- args passing
(+) better diffing because yaml lockfile

(-) more configuraiton
(-) needs to be installed
(-) tarball (requires config)

**Vote:**

👍 We want PNPM as the new package manager
👎 Stick to the pain we know for all the NPM lovers
 

@siemens/siemens-element-members Please share your voice on the package manager topic!

kfenner:
> See #1900
> 
> **Summary:**
> 
> (+) way more powefull than npm (+) more security (+) releasing is less complex: no postversion (+) stable lockfile (+) no -- args passing (+) better diffing because yaml lockfile
> 
> (-) more configuraiton (-) needs to be installed (-) tarball (requires config)
> 
> **Vote:**
> 
> 👍 We want PNPM as the new package manager 👎 Stick to the pain we know for all the NPM lovers  
> 
> @siemens/siemens-element-members Please share your voice on the package manager topic!

@spliffone I would say the voting is complete and the majority is positive about switching to pnpm. Should we do this migration for Element v50 so we stick to npm here on the main branch until we start merging stuff for v50 and v49.x also stays on npm, OK?

.npmrc
```diff
@@ -1,3 +1,3 @@
-@simpl:registry=https://code.siemens.com/api/v4/packages/npm/
+@simpl:registry=https://code.siemens.com/api/v4/projects/732/packages/npm/
```
spike-rabbit, line 1: do we really need this change?

package.json
```diff
@@ -13,10 +13,9 @@
     "email": "info@siemens.com"
   },
   "type": "module",
+  "packageManager": "pnpm@11.9.0",
   "engines": {
-    "node": "^24.0.0",
-    "yarn": "Please use NPM to install dependencies",
-    "npm": "11"
+    "node": "^24.0.0"
```
spike-rabbit, line 18: I think you should put pnpm here, and explicitly block yarn and npm

pnpm-workspace.yaml
```diff
@@ -0,0 +1,97 @@
...
+  - qrcode-generator
+  - react
+  - react-dom
+  - vue
+  - vue3-sfc-loader
+
+# Suppresses errors for unmet peer deps from third-party packages that have not
+# updated their ranges yet (e.g. @angular-architects/module-federation requiring
+# @angular/core@^21.2.0 while the workspace pins 21.0.8).
+strictPeerDependencies: false
```
spike-rabbit: Cannot we have dependency overrides again? Otherwise we loosen checks compared to before

spike-rabbit (changes requested):
Looks already pretty good. Just some hopefully minor stuff.

.npmrc
spliffone: Sadly it was necessary since pnpm build the wrong download tarball url. I guess we can also fix it later, I would also like to get rid of the tarball urls in the lockfile but this cause the same issue.

projects/dashboards-demo/federation.config.mjs
```diff
@@ -25,7 +25,13 @@ export default withNativeFederation({
...
     /^echarts/,
     // Skip packages with assets that can't be bundled for browser
-    'flag-icons'
+    'flag-icons',
+    // Skip Node.js-only build/schematics tooling that shareAll would otherwise
+    // try to bundle for the browser. These transitively import
+    // @angular-devkit/core, whose template.js is not valid browser code and
+    // breaks esbuild ("Unterminated string literal").
+    /^@angular-architects\/module-federation/,
+    /^@angular\/cdk\/schematics/
```
chintankavathia: is this needed?

package.json
```diff
@@ -186,22 +173,11 @@
     "vitest": "4.1.9"
   },
   "optionalDependencies": {
-    "@siemens-ux/design-tokens": "^0.5.0",
+    "@siemens-ux/design-tokens": "0.6.0",
```
chintankavathia, line 176: is this bump necessary for pnpm migration?

spliffone: No, I can roll it back

package.json
spliffone: I wouldn't change this, this seems to be the standard for angular projects to define the compatible node version. But I replace the `packageManager` by:

```json
"devEngines": {
    "packageManager": {
      "name": "pnpm",
      "version": ">=11.13.0",
      "onFail": "download"
    }
  }
```

spike-rabbit (approved):
👍

PR #1949, opened 2026-04-23
feat(list-item): introduce new styles

### LINKED ISSUE #1641: Transform notification item in a generic list item
### Prerequisites

### Feature request

The notification needs to be  refactored into a list component. 

During the design process it became clear that the structure and behavior are not specific to notifications but represent a generic list item pattern that can be reused across different contexts. Defining it as a list component allows the same pattern to support notifications as well as other list-based content.

The notification variant builds on the same base structure, with the addition of:

-  Metadata
- A different placement for the main action
- Padding adjustments
- Quick actions have been changed by default to a button group

The following aspects remain unchanged:

- Interaction states
- Status
- Overflow behaviors
- Overarching layout

Please reach out for Figma specs

### Use case

_No response_

ljanner:
Resolves #1641

---

[Documentation](https://d33c9dcnqinn2a.cloudfront.net/pr-1949/pages/).
[Examples](https://d33c9dcnqinn2a.cloudfront.net/pr-1949/pages/element-examples/).
[Dashboards Demo](https://d33c9dcnqinn2a.cloudfront.net/pr-1949/pages/dashboards-demo/).
[Playwright report](https://d33c9dcnqinn2a.cloudfront.net/pr-1949/playwright-report/).

**Coverage Reports:**

Code Coverage

- [charts-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-1949/pages/coverage/charts-ng/)
- [dashboards-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-1949/pages/coverage/dashboards-ng/)
- [element-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-1949/pages/coverage/element-ng/)
- [element-translate-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-1949/pages/coverage/element-translate-ng/)
- [maps-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-1949/pages/coverage/maps-ng/)
- [native-charts-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-1949/pages/coverage/native-charts-ng/)


ljanner:
@spike-rabbit could you take a look at the example and provide some initial feedback if the structure looks as expected?

projects/element-theme/src/styles/components/_list-item.scss
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+  .list-item-timestamp {
+    color: variables.$element-text-secondary;
+    margin-block-end: 0;
+  }
+
+  .list-item-title {
+    grid-column: 1;
+    block-size: map.get(variables.$spacers, 9);
```
dr-itz: use something that grows with bigger root font-size.  and better avoid block-size when possible. here you instead want


```suggestion
    padding-block: map.get(variables.$spacers, 4);
```

i.e. 8px. with the line-height of 16px this gives the same 32px but grows correctly with bigger root font size

projects/element-theme/src/styles/components/_list-item.scss
```diff
@@ -0,0 +1,91 @@
...
+    margin-block-end: 0;
+
+    @include typography.si-font(
+      typography.$si-font-size-h5,
+      typography.$si-line-height-h5,
+      typography.$si-font-weight-h5
+    );
+  }
+
+  .list-item-primaryaction {
```
chintankavathia: nit

```suggestion
  .list-item-primary-action {
```

projects/element-theme/src/styles/components/_list-item.scss
```diff
@@ -0,0 +1,91 @@
...
+    .list-item-metadata-divider {
+      inline-size: 0.125rem;
+      block-size: 0.125rem;
+      border-radius: 50%;
+      background-color: variables.$element-text-secondary;
+      margin-inline: map.get(variables.$spacers, 2);
+    }
+  }
+
+  .list-item-quickactions {
```
chintankavathia:
```suggestion
  .list-item-quick-actions {
```

chintankavathia (dismissed):

panch1739 (commented):
@ljanner Looks great!

Just one observation...in the list-item-actions, the active/selected state seems to be missing. When the list item is pressed, the background color should change to base-1-selected. 

Questions:

- The replacement of the notification item would happen in another PR?
- Does this have the "unread" property also? (This is only valid for notification item)

projects/element-theme/src/styles/components/_list-item.scss
ljanner: applied 👍

projects/element-theme/src/styles/components/_list-item.scss
ljanner: applied 👍

projects/element-theme/src/styles/components/_list-item.scss
ljanner: applied 👍

ljanner:
@panch1739 
> * Does this have the "unread" property also? (This is only valid for notification item)

I now added a new example called list-item-unread. While an unread state isn't quite generic I do think it makes sense to provide it since the idea is to drop the notification item entirely.  

> The replacement of the notification item would happen in another PR?

For the replacement itself, we need to choose a technical solution either:
- Deprecate the notification item directly in this PR
- Rebuild it internally to use list-item (though the effort may not be worth it)

This is still open, we can see how far we can get with migrations for element v50.

 For now I added a banner on the notification item docs recommending list-item instead.

panch1739:
> @panch1739
> 
> > * Does this have the "unread" property also? (This is only valid for notification item)
> 
> I now added a new example called list-item-unread. While an unread state isn't quite generic I do think it makes sense to provide it since the idea is to drop the notification item entirely.
> 
> > The replacement of the notification item would happen in another PR?
> 
> For the replacement itself, we need to choose a technical solution either:
> 
> * Deprecate the notification item directly in this PR
> * Rebuild it internally to use list-item (though the effort may not be worth it)
> 
> This is still open, we can see how far we can get with migrations for element v50.
> 
> For now I added a banner on the notification item docs recommending list-item instead.

@ljanner Excellent! Is good for me :) Thank you very much!

src/app/examples/list-item/list-item-action.html
```diff
@@ -0,0 +1,33 @@
+<div class="list-group">
```
spliffone: We should use `<ul>` + `<li>` elements to ensure the example has a list semantic

src/app/examples/list-item/list-item-action.html
```diff
@@ -0,0 +1,33 @@
...
+  <div class="list-item list-item-action" (click)="logEvent('Open project Alpha')">
```
spliffone: Elements which are clickable should be either buttons or anchors

src/app/examples/list-item/list-item-action.html
```diff
@@ -0,0 +1,33 @@
...
+    <p class="list-item-timestamp">Edited 2 hours ago</p>
```
spliffone: Semantical correct is the [time](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/time) element instead of a paragraph

src/app/examples/list-item/list-item-action.html
```diff
@@ -0,0 +1,33 @@
...
+        <si-icon class="icon-sm" [icon]="icons.elementUser" />
+        <span>3 contributors</span>
+      </div>
+    </div>
+  </div>
+  <div class="list-item list-item-action" (click)="logEvent('Open project Gamma')">
+    <p class="list-item-timestamp">Edited last week</p>
+    <h5 class="list-item-title">Project Gamma</h5>
+    <p class="list-item-description">Safety interlock logic for press station</p>
+    <div class="list-item-metadata">
```
spliffone: The purpose of .list-item-metadata isn't clear, we should describe it

src/app/examples/list-item/list-item-action.html
```diff
@@ -0,0 +1,33 @@
+<div class="list-group">
+  <div class="list-item list-item-action" (click)="logEvent('Open project Alpha')">
+    <p class="list-item-timestamp">Edited 2 hours ago</p>
+    <h5 class="list-item-title">Project Alpha</h5>
+    <p class="list-item-description">Automation config for assembly line 3</p>
+  </div>
+  <div class="list-item list-item-action" (click)="logEvent('Open project Beta')">
+    <p class="list-item-timestamp">Edited 3 days ago</p>
+    <h5 class="list-item-title">Project Beta</h5>
```
spliffone: Optionally, we could link the title and description together via aria-describedby

src/app/examples/list-item/list-item-action.html
ljanner: Done 👍

src/app/examples/list-item/list-item-action.html
ljanner: Done 👍

src/app/examples/list-item/list-item-action.html
ljanner: Added docs 👍

src/app/examples/list-item/list-item-action.html
ljanner: Described it in the docs and did it in the example too 👍

src/app/examples/list-item/list-item-action.html
ljanner: Done 👍

projects/element-theme/src/styles/components/_list-item.scss
```diff
@@ -0,0 +1,134 @@
...
+      &::before {
+        content: '';
+        position: absolute;
+        block-size: map.get(variables.$spacers, 3);
+        inline-size: map.get(variables.$spacers, 3);
+        border-radius: 50%;
+        background-color: variables.$element-ui-1;
+        inset-inline-start: calc(
+          -1 * (map.get(variables.$spacers, 2) + map.get(variables.$spacers, 3))
+        );
```
dr-itz: no need for a `calc()`, can be calculated at compile time since it's all just px values

projects/element-theme/src/styles/components/_list-item.scss
```diff
@@ -0,0 +1,134 @@
...
+        typography.$si-font-size-h5-bold,
+        typography.$si-line-height-h5-bold,
+        typography.$si-font-weight-h5-bold
+      );
+
+      &::before {
+        content: '';
+        position: absolute;
+        block-size: map.get(variables.$spacers, 3);
+        inline-size: map.get(variables.$spacers, 3);
```
dr-itz: should this be fixed size or grow with bigger font? i.e. 0.375rem instead?

projects/element-theme/src/styles/components/_list-item.scss
```diff
@@ -0,0 +1,134 @@
...
+      color: variables.$element-text-secondary;
+    }
+
+    .badge {
+      margin-inline: 0;
+    }
+
+    .list-item-metadata-divider {
+      inline-size: 0.125rem;
+      block-size: 0.125rem;
```
dr-itz, line 95: ..this one is in rem, so the one above should be consistently also in rem. or both in px

projects/element-theme/src/styles/components/_list-item.scss
ljanner: Changed to rem 👍

projects/element-theme/src/styles/components/_list-item.scss
ljanner: Changed to rem 👍

projects/element-theme/src/styles/components/_list-item.scss
ljanner: Dropped calc but using rem for proper scaling.

projects/element-theme/src/styles/components/_list-item.scss
```diff
@@ -0,0 +1,132 @@
...
+    }
+  }
+
+  &:is(button, a) {
+    border: 0;
+    background: none;
+    inline-size: 100%;
+    text-align: start;
+    font: inherit;
+  }
```
spike-rabbit, line 131: is there a need to only scope this to button/a?

src/app/examples/list-item/list-item-action.html
```diff
@@ -0,0 +1,57 @@
+<ul>
+  <li>
+    <button
+      type="button"
+      class="list-item list-item-action"
+      aria-labelledby="alpha-title"
+      aria-describedby="alpha-desc"
```
spike-rabbit: Did you check this with the a11y team?
I guess having the entire content as description might be correct.

spike-rabbit (dismissed):

projects/element-theme/src/styles/components/_list-item.scss
ljanner: The reason why I scoped it to `button` and `a` is because those are the expected intractable elements used as a `list-item-action`.

src/app/examples/list-item/list-item-action.html
ljanner: Sent out a message to the a11y team, wdyt can we proceed in the mean time or shall we wait for their answer?

src/app/examples/list-item/list-item-unread.ts
```diff
@@ -0,0 +1,17 @@
...
+import { SiFormItemComponent } from '@siemens/element-ng/form';
+
+@Component({
+  selector: 'app-sample',
+  imports: [FormsModule, SiFormItemComponent],
+  templateUrl: './list-item-unread.html',
+  changeDetection: ChangeDetectionStrategy.OnPush
+})
+export class SampleComponent {
+  unread = true;
```
spliffone: Can we use a signal to be onpush compatibility?

```suggestion
  readonly unread = signal(true);
```

src/app/examples/list-item/list-item.html
```diff
@@ -0,0 +1,108 @@
...
+    <div class="list-item-primary-action">
+      <button
+        type="button"
+        class="btn btn-icon btn-ghost element-options-vertical"
+        aria-label="Options"
+        [cdkMenuTriggerFor]="contextMenu"
+      ></button>
+    </div>
+    <p class="list-item-description"
+      >SIMATIC S7-1500 firmware v3.1.2 is ready for deployment. Review and approve to proceed with
```
spliffone: I wouldn't explicitly mention Siemens products

src/app/examples/list-item/list-item-unread.html
```diff
@@ -0,0 +1,40 @@
...
+  <li class="list-item">
+    <time class="list-item-timestamp" datetime="11:45">Today at 11:45</time>
+    <h5 class="list-item-title" [class.unread]="unread">Sensor calibration report available</h5>
+    <p class="list-item-description"
+      >Calibration of temperature sensors on production line 2 has been completed and documented.</p
+    >
+    <div class="list-item-metadata">
+      <span>Production line 2</span>
+      <div class="list-item-metadata-divider"></div>
+      <a href="#">View report</a>
```
spliffone, line 20: The link is really hard distinguish from the other text elements

@panch1739 should we underline the link?

src/app/examples/list-item/list-item.html
ljanner: Removed product name 👍

src/app/examples/list-item/list-item-unread.ts
ljanner: Applied 👍

src/app/examples/list-item/list-item.ts
```diff
@@ -0,0 +1,37 @@
...
+} from '@siemens/element-icons';
+import { addIcons, SiIconComponent } from '@siemens/element-ng/icon';
+import { SiMenuFactoryComponent, type MenuItem } from '@siemens/element-ng/menu';
+import { LOG_EVENT } from '@siemens/live-preview';
+
+@Component({
+  selector: 'app-sample',
+  imports: [SiIconComponent, SiMenuFactoryComponent, CdkMenuTrigger],
+  templateUrl: './list-item.html',
+  changeDetection: ChangeDetectionStrategy.OnPush
```
spliffone: Does it mal sense to add a padding around?

```suggestion
  changeDetection: ChangeDetectionStrategy.OnPush,
  host: { class: 'p-5' }
```

src/app/examples/list-item/list-item-unread.ts
```diff
@@ -0,0 +1,17 @@
...
+ */
+import { ChangeDetectionStrategy, Component, signal } from '@angular/core';
+import { FormsModule } from '@angular/forms';
+import { SiFormItemComponent } from '@siemens/element-ng/form';
+
+@Component({
+  selector: 'app-sample',
+  imports: [FormsModule, SiFormItemComponent],
+  templateUrl: './list-item-unread.html',
+  changeDetection: ChangeDetectionStrategy.OnPush
```
spliffone: Should we add a padding?

src/app/examples/list-item/list-item-action.ts
```diff
@@ -0,0 +1,21 @@
...
+import { elementUser, elementLock } from '@siemens/element-icons';
+import { addIcons, SiIconComponent } from '@siemens/element-ng/icon';
+import { LOG_EVENT } from '@siemens/live-preview';
+
+@Component({
+  selector: 'app-sample',
+  imports: [SiIconComponent],
+  templateUrl: './list-item-action.html',
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  host: { class: 'p-4' }
```
spliffone: Usually we use p-5 as padding in our examples

src/app/examples/list-item/list-item-unread.html
panch1739: @spliffone I cannot fully understand the context just looking at the HTML , but it depends on how is surrounded by text:

- Links in paragraphs should always be underlined to ensure they’re easily recognized. As standalone elements (e.g., in cards or lists), underlines aren’t needed and an arrow icon (>) can be used to clarify function.
- If a link needs to trigger an action instead of navigating (e.g., opening a modal or expanding content), use a button link instead.

src/app/examples/list-item/list-item-action.html
ljanner: Had a call with two colleagues from a11y team. No `aria-*by` should be set, otherwise information is lost.

src/app/examples/list-item/list-item.html
```diff
@@ -0,0 +1,107 @@
...
+      <span>Production line 2</span>
+      <div class="list-item-metadata-divider"></div>
+      <a href="#">View report</a>
+    </div>
+  </li>
+
+  <li class="list-item">
+    <time class="list-item-timestamp" datetime="09:30">Today at 09:30</time>
+    <h5 class="list-item-title">Firmware update pending approval</h5>
+    <div class="list-item-primary-action">
```
spike-rabbit, line 27: I am not exactly in favor of the name `primary-action`. 
The UX specs only show a menu there. So what about calling it `list-item-menu`?

spike-rabbit (changes requested):

src/app/examples/list-item/list-item.html
spliffone: I would go something like context-action, the position is at the end and its doesn't need to be a menu it could also be on or more action buttons.

Just a side question, should we shorten the naming?
.list-item:
- .title
- .timestamp
- .description  
- .metadata

ljanner: The naming here is difficult and has previously also been a discussion for the notification-item. Here is the link to the full discussion (internal): https://code.siemens.com/simpl/simpl-element/-/work_items/2371#note_39626788

I also think `list-item-menu` is too specific as other action buttons are totally valid. To me personally the name `context-action` lacks hierarchal meaning. 

I'm open to something like `list-item-main-action` or `list-item-inline-action`. 

/cc @kfenner

spike-rabbit: I see, then I am fine with primary-action as we had it before.

spike-rabbit (approved):
👍

spike-rabbit (approved):

PR #1956, opened 2026-04-24
refactor(select): use model signal for `value`

spike-rabbit:
The `si-select` selection strategy was still using an `@Input`. This replaces it with a `model`.
As a side effect `valueChange` can now emit undefined. This is actually more correct than the previous approach.

BREAKING CHANGE:
`si-select` is now using a `model` for the `value(Change)` input / output. With that, the type now indicates that undefined can be emitted as well. This may require adjustments when using the `valueChange` event.


spike-rabbit:
does not work

PR #1968, opened 2026-04-28
ci: post AI fix prompt on PR when pipeline fails

chintankavathia:


chintankavathia:
not needed. one can directly use PR # with github cli

PR #1982, opened 2026-04-29
fix(status-toggle): use space key to activate radio items

akashsonune:
https://www.w3.org/WAI/ARIA/apg/patterns/radio/

---

[Documentation](https://d33c9dcnqinn2a.cloudfront.net/pr-1982/pages/).
[Examples](https://d33c9dcnqinn2a.cloudfront.net/pr-1982/pages/element-examples/).
[Dashboards Demo](https://d33c9dcnqinn2a.cloudfront.net/pr-1982/pages/dashboards-demo/).
[Playwright report](https://d33c9dcnqinn2a.cloudfront.net/pr-1982/playwright-report/).

**Coverage Reports:**

Code Coverage

- [charts-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-1982/pages/coverage/charts-ng/)
- [dashboards-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-1982/pages/coverage/dashboards-ng/)
- [element-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-1982/pages/coverage/element-ng/)
- [element-translate-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-1982/pages/coverage/element-translate-ng/)
- [maps-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-1982/pages/coverage/maps-ng/)
- [native-charts-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-1982/pages/coverage/native-charts-ng/)

### DIFF
--- a/projects/element-ng/status-toggle/si-status-toggle.component.html
+++ b/projects/element-ng/status-toggle/si-status-toggle.component.html
@@ -25,12 +25,17 @@
       class="status-toggle-item focus-inside px-4"
       role="radio"
       [attr.aria-checked]="$index === selIndexValue"
-      [attr.tabindex]="isDisabled() || item.disabled ? '' : '0'"
+      [attr.tabindex]="
+        isDisabled() || item.disabled ? null : $index === focusedIndex() ? '0' : '-1'
+      "
       [attr.aria-disabled]="isDisabled() || item.disabled"
       [class.active]="$index === selIndexValue"
       [class.disabled]="item.disabled || isDisabled()"
       [style.width.%]="100 / $count"
+      (keydown.space)="$event.preventDefault(); selectItem($index)"
       (keydown.enter)="selectItem($index)"
+      (keydown.arrowRight)="navigateFocus($event, $index, 1)"
+      (keydown.arrowLeft)="navigateFocus($event, $index, -1)"
     >
       <span class="icon icon-stack">
         <si-icon
--- a/projects/element-ng/status-toggle/si-status-toggle.component.ts
+++ b/projects/element-ng/status-toggle/si-status-toggle.component.ts
@@ -16,7 +16,8 @@ import {
   output,
   signal,
   SimpleChanges,
-  viewChild
+  viewChild,
+  viewChildren
 } from '@angular/core';
 import { ControlValueAccessor, NG_VALUE_ACCESSOR } from '@angular/forms';
 import { isRTL, listenGlobal } from '@siemens/element-ng/common';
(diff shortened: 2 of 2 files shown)


dr-itz:
I think these need to go from tab-activated to arrow left/right. i.e. the active one should have tabindex=0, the others tabindex=-1 and arrow/left (with RTL compensation) should move the focus and active the element.

But. In some use-cases changing the status triggers a potential heavy operation, so an explicit activation with space is probably more desirable but then we also need to wrap the the whole thing into a role=toolbar (i.e. on the host itself)

akashsonune:
> I think these need to go from tab-activated to arrow left/right. i.e. the active one should have tabindex=0, the others tabindex=-1 and arrow/left (with RTL compensation) should move the focus and active the element.
> 
> But. In some use-cases changing the status triggers a potential heavy operation, so an explicit activation with space is probably more desirable but then we also need to wrap the the whole thing into a role=toolbar (i.e. on the host itself)

Done. I think it also mentions with the toolbar, the arrow navigation should be move around i.e. from last element to the first and first to last(back arrow)

akashsonune:
Also I noticed that, selecting the item with mouse click does not focus it. so when we use keyboard later on , the focus is elsewhere while the selection is different. this was also the case earlier with tab navigation

dr-itz:
The problem I see now is that we get a focus border when operated with the mouse. This shouldn't happen

akashsonune:
> The problem I see now is that we get a focus border when operated with the mouse. This shouldn't happen

@dr-itz That was intentional. As I mentioned above,  the focus will be elsewhere , when user tries to use keyboard navigation.
I see similar pattern at other examples- https://carbondesignsystem.com/components/content-switcher/usage/
also https://www.w3.org/WAI/ARIA/apg/patterns/radio/examples/radio/

dr-itz:
> > The problem I see now is that we get a focus border when operated with the mouse. This shouldn't happen
> 
> @dr-itz That was intentional. As I mentioned above, the focus will be elsewhere , when user tries to use keyboard navigation. I see similar pattern at other examples- https://carbondesignsystem.com/components/content-switcher/usage/ also https://www.w3.org/WAI/ARIA/apg/patterns/radio/examples/radio/

Yes, but the visual focus ring isn't expected when using the mouse. Other controls don't do that

projects/element-ng/status-toggle/si-status-toggle.component.html
spike-rabbit, line 23: why not using a proper input with a label? Then a lot bindings are not needed?

spike-rabbit (changes requested):

PR #2034, opened 2026-05-11
Restore tooltips on si-navbar-vertical and si-content-action-bar

fabrice-aeschbacher:
MenuItem from @simpl/element-ng/menu has been deprecated in v47.0.0, but the new implementation removed the tooltips. This patch restores the feature, while fully respecting the new implementation in case no optional tooltip is given.


spliffone:
@panch1739 Can you have a look here, if we want to have a tooltip in the components I can support Fabrice here

PR #2077, opened 2026-05-18
feat(tabs): make menu keyboard reachable

akashsonune:


akashsonune:
In the tabs-menu example -https://element.siemens.io/element-examples/#/overview/si-tabs/si-tabs-menu?q=tabs
When the menu is visible, it is not reachable via keyboard,

I do not see any Accessibility issue since we can reach all the tabs via keyboard arrows, but having the menu button in tab order would improve user experience in my opinion, as It would help to visualize all tabs at once in such case.

@panch1739 @dr-itz  WDYT?

spike-rabbit:
Just discussed this with the a11y, they tihnk we should keep it hidden as is, since for screenreader this would be useless tab stop.

akashsonune:
> Just discussed this with the a11y, they tihnk we should keep it hidden as is, since for screenreader this would be useless tab stop.

Alright 👍

PR #2106, opened 2026-05-28
feat(tree-view): add copy-drag directive for non-destructive drag operations

TaylorShane:
## Description
Adds a new `SiCopyDragDirective` that enables copy-drag for tree-view items. When applied alongside cdkDrag, the source item remains visible during the drag operation instead of being hidden by CDK's default placeholder behavior and remains in the sources tree to be dragged again to a new location.

This is useful for scenarios like dragging items from a catalog/palette into a working tree without removing them from the source. This could be used to assign devices to locations or creating an entire building hierarchy via drag and drop.

## Usage
Pair with transferTreeItem(..., { removeFromSource: false }) in the drop handler to preserve the source data model.

## Changes
New: `SiCopyDragDirective` — standalone directive that clones the CDK placeholder so the original item stays visible during drag
New: `si-tree-view-drag-drop-copy example` — demo with two side-by-side trees (building tree + equipment catalog) showing copy-drag in action
New: 7 unit tests covering the directive and transferTreeItem copy semantics (source preservation, deep clone independence, child/sibling insertion)
Export: Added `SiCopyDragDirective` to the tree-view public API

## How it works
The directive subscribes to CDK Drag's `started`/`ended` events. On drag start, it clones the placeholder element (which CDK inserts to replace the dragged item), removes the `cdk-drag-placeholder` class so it isn't hidden by existing CSS, and inserts it after the placeholder. On drag end, the clone is removed.

## Testing
16/16 tree-view tests pass (9 existing + 7 new)
No changes to existing APIs or behavior

tree-drag-and-drop-copy

### DIFF
--- a/projects/element-ng/tree-view/si-copy-drag.directive.ts
+++ b/projects/element-ng/tree-view/si-copy-drag.directive.ts
@@ -1,0 +1,50 @@
+/**
+ * Copyright (c) Siemens 2016 - 2026
+ * SPDX-License-Identifier: MIT
+ */
+import { CdkDrag } from '@angular/cdk/drag-drop';
+import { DestroyRef, Directive, inject } from '@angular/core';
+import { takeUntilDestroyed } from '@angular/core/rxjs-interop';
+
+/**
+ * Add alongside `cdkDrag` to keep the source element visible during drag,
+ * providing copy-drag semantics for tree-view items.
+ *
+ * The directive clones the CDK placeholder so the original item remains
+ * visible while being dragged to another tree.
+ *
+ * @example
+ * ```html
+ * <si-tree-view-item cdkDrag siCopyDrag [cdkDragData]="item">
+ *   <si-tree-view-item *cdkDragPreview />
+ * </si-tree-view-item>
+ * ```
+ */
+@Directive({
+  selector: '[siCopyDrag]',
+  host: {
+    class: 'si-copy-drag'
+  }
+})
+export class SiCopyDragDirective {
+  private clone: HTMLElement | null = null;
+
+  constructor() {
+    const drag = inject(CdkDrag);
+    const destroyRef = inject(DestroyRef);
+
+    drag.started.pipe(takeUntilDestroyed(destroyRef)).subscribe(() => {
+      const placeholder = drag.getPlaceholderElement();
+      this.clone = placeholder.cloneNode(true) as HTMLElement;
+      this.clone.classList.remove('cdk-drag-placeholder');
+      this.clone.style.pointerEvents = 'none';
+      this.clone.removeAttribute('id');
+      placeholder.after(this.clone);
+    });
+
+    drag.ended.pipe(takeUntilDestroyed(destroyRef)).subscribe(() => {
+      this.clone?.remove();
+      this.clone = null;
+    });
+  }
+}
--- a/projects/element-ng/tree-view/si-copy-drag.directive.ts
+++ b/projects/element-ng/tree-view/si-copy-drag.directive.ts
@@ -33,6 +33,10 @@ export class SiCopyDragDirective {
     const drag = inject(CdkDrag);
     const destroyRef = inject(DestroyRef);
 
+    destroyRef.onDestroy(() => {
+      this.clone?.remove();
+    });
+
     drag.started.pipe(takeUntilDestroyed(destroyRef)).subscribe(() => {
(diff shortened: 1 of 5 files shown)


projects/element-ng/tree-view/si-tree-view-item/si-tree-view-item.component.spec.ts
```diff
@@ -256,3 +257,213 @@ describe('SiTreeViewComponentWithDragDrop', () => {
...
+    ]);
+  }
+
+  it('renders items with siCopyDrag directive', async () => {
+    const fixture = TestBed.createComponent(CopyDragWrapperComponent);
+    await fixture.whenStable();
+
+    const items = fixture.debugElement.queryAll(By.css('si-tree-view-item'));
+    expect(items).toHaveLength(2);
+    expect(items[0].nativeElement.classList).toContain('si-copy-drag');
```
spliffone, line 298:
```suggestion
    expect(items[0].nativeElement).toHaveClass('si-copy-drag');
```

projects/element-ng/tree-view/si-tree-view-item/si-tree-view-item.component.spec.ts
```diff
@@ -256,3 +257,213 @@ describe('SiTreeViewComponentWithDragDrop', () => {
...
+    const target: TreeItem = {
+      label: 'Floor 1',
+      state: 'expanded',
+      children: [{ label: 'Room 101', state: 'leaf' }]
+    };
+    const targetItems = [target];
+    initTree(targetItems);
+
+    const targetFlat = flatten(targetItems);
+    const event = makeDrop(sourceItems, targetFlat, 0, 1);
```
spliffone, line 351: Can you please make use of the userEvent API (see https://main.vitest.dev/api/browser/interactivity.html#userevent-draganddrop) for all the test cases

src/app/examples/si-tree-view/si-tree-view-drag-drop-copy.html
```diff
@@ -0,0 +1,65 @@
+<h3>Copy between trees</h3>
+<p
+  >Drag equipment from the catalog into a room. Items are copied (not moved) and inserted as
```
spliffone, line 3: The descriptions seems to be specfic for the buildings domain, can you come up with a more general use case?

src/app/examples/si-tree-view/si-tree-view-drag-drop-copy.ts
```diff
@@ -0,0 +1,87 @@
...
+            { label: 'Room 102', icon: 'element-room', state: 'leaf' }
+          ]
+        }
+      ]
+    }
+  ];
+
+  equipmentItems: TreeItem[] = [
+    {
+      label: 'PXC5.E003',
```
spliffone, line 52: Try to prevent building specific terms and attribute names. Please focus on a general use case and terminology.

projects/element-ng/tree-view/si-copy-drag.directive.ts
```diff
@@ -0,0 +1,54 @@
...
+ * </si-tree-view-item>
+ * ```
+ */
+@Directive({
+  selector: '[siCopyDrag]',
+  host: {
+    class: 'si-copy-drag'
+  }
+})
+export class SiCopyDragDirective {
```
spliffone, line 29: @chintankavathia What do you think of the name? I would prefer to have `tree` in the name.

projects/element-ng/tree-view/si-copy-drag.directive.ts
```diff
@@ -0,0 +1,54 @@
...
+  private clone: HTMLElement | null = null;
+
+  constructor() {
+    const drag = inject(CdkDrag);
```
spliffone, line 33: Without `CDKDrag` the directive will fail, I guess you have 3 options:
- use hostDirective
-  combine the selector [siCopyDrag][cdkDrag]
- or add a debug mode warning and make the CDK directive optional

spliffone (changes requested):

chintankavathia:
@TaylorShane  Thanks for the PR but this custom directive is not needed as CDK already has `cdkDropListHasAnchor` https://angular.dev/guide/drag-drop#copying-items-between-lists 
currently our custom styles are not honoring it. I will adjust some styles so that `cdkDropListHasAnchor` can be used.

TaylorShane:
functionality provided in #2110

PR #2175, opened 2026-06-17
feat: provide custom builder with esbuild plugins

spliffone:

### DIFF
--- a/projects/live-preview/builders/src/application/schema.json
+++ b/projects/live-preview/builders/src/application/schema.json
@@ -1,0 +1,738 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema",
+  "title": "Application builder with esbuild plugin support",
+  "description": "Application builder with custom esbuild plugin support. Accepts all options from @angular/build:application.",
+  "type": "object",
+  "properties": {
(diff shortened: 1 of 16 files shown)


projects/live-preview/builders/src/application/schema.json
```diff
@@ -0,0 +1,799 @@
...
+  "type": "object",
+  "properties": {
+    "plugins": {
+      "type": "array",
+      "description": "List of paths to esbuild plugins. Each path should resolve to a module that exports an esbuild Plugin or Plugin array.",
+      "items": {
+        "type": "string"
+      },
+      "default": []
+    },
```
michael-smt, line 14: As far as I understand this is a copy of the Angular builder's schema extended with this section. Generating this file might be better, so we don't have to remember to keep this file up-to-date with Angular changes.

projects/live-preview/builders/src/dev-server/schema.json
michael-smt, line 1: Same here as with the application schema, I think this would be better if copied at builder's build time.

michael-smt (commented):
I might be out of the loop, but what's the reason for this? 🙂

PR #2314, opened 2026-07-10
refactor(split): provide collapse label translation key and use svg icon

spliffone:
---

[Documentation](https://d33c9dcnqinn2a.cloudfront.net/pr-2314/pages/).
[Examples](https://d33c9dcnqinn2a.cloudfront.net/pr-2314/pages/element-examples/).
[Dashboards Demo](https://d33c9dcnqinn2a.cloudfront.net/pr-2314/pages/dashboards-demo/).
[Playwright report](https://d33c9dcnqinn2a.cloudfront.net/pr-2314/playwright-report/).

**Coverage Reports:**

Code Coverage

- [charts-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-2314/pages/coverage/charts-ng/)
- [dashboards-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-2314/pages/coverage/dashboards-ng/)
- [element-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-2314/pages/coverage/element-ng/)
- [element-translate-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-2314/pages/coverage/element-translate-ng/)
- [maps-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-2314/pages/coverage/maps-ng/)
- [native-charts-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-2314/pages/coverage/native-charts-ng/)


projects/element-ng/split/si-split-part.component.ts
```diff
@@ -51,10 +53,8 @@ export class SiSplitPartComponent {
   /**
    * Sets the icon class that is used in the buttons of split parts to
    * collapse and uncollapse the parts.
-   *
-   * @defaultValue 'element-double-right'
    */
-  readonly collapseIconClass = input('element-double-right');
+  readonly collapseIconClass = input<string | undefined>();
```
spike-rabbit, line 57: Why not


```suggestion
  readonly collapseIconClass = input(this.icons.elementDoubleRight);
```

? This would produce a meaningful default value and we don't need to have undefined in the type

spike-rabbit (changes requested):

projects/element-ng/split/si-split-part.component.ts
spliffone: I was thinking the same, but once assigned by the user we have no way the fallback. Further, I guess the default value `this.icons.elementDoubleRight` isn't really helpful for consumers.

So you still want the change?

spike-rabbit: What about just keeping `@defaultValue 'element-double-right'` + your code changes?

spliffone: We can but so far the SVGs were used as default icons, so if we should discuss it in the core team.

spike-rabbit: after thinking about this again, why is this even an input at all? I don't think consumers should change this

spliffone: I seem we had some demand for it since we have an explicit example `si-split/si-split-icons` and the input was available from the beginning. Further, we usually allow to customize the icons everywhere so we is this an exception?

spike-rabbit: I see those arrows more like the dropdown arrow we have for instance in dropdowns. There we also do not allow overriding it. But let me check this with UX. But I guess you are anyway right that we should not just remove it.

spike-rabbit (approved):

PR #2386, opened 2026-07-21
chore(docs): use zensical

Killusions:
Switch to more modern docs-tooling (and use the currently only viable way of modifying the config with a "plugin")

---

[Documentation](https://d33c9dcnqinn2a.cloudfront.net/pr-2386/pages/).
[Examples](https://d33c9dcnqinn2a.cloudfront.net/pr-2386/pages/element-examples/).
[Dashboards Demo](https://d33c9dcnqinn2a.cloudfront.net/pr-2386/pages/dashboards-demo/).
[Playwright report](https://d33c9dcnqinn2a.cloudfront.net/pr-2386/playwright-report/).

**Coverage Reports:**

Code Coverage

- [charts-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-2386/pages/coverage/charts-ng/)
- [dashboards-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-2386/pages/coverage/dashboards-ng/)
- [element-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-2386/pages/coverage/element-ng/)
- [element-translate-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-2386/pages/coverage/element-translate-ng/)
- [maps-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-2386/pages/coverage/maps-ng/)
- [native-charts-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-2386/pages/coverage/native-charts-ng/)


Killusions:
@spike-rabbit @kfenner @spliffone Already prepared this using the same approach we use in the docs-theme (and docs-composer MR, not needed for this though). This is all that's needed and the output is the exactly same (except for the search, which is better). Whenever you're ready to migrate to enjoy the faster build speeds and more stability, this is ready.

projects/element-docs-builder/md_extension_element_docs/extension.py
```diff
@@ -105,19 +105,32 @@ def __init__(self, examples_base: str, page: str, *args: Any, **kwargs: Any) ->
                     for correct URL construction (e.g., '/element-examples/').
       page: Current page's absolute URL from MkDocs context.
     """
-    self.examples_base = examples_base
+    self.configured_examples_base = examples_base
     self.page = page
+    self.set_examples_base()
+    super().__init__('si-docs-component', *args, **kwargs)
+
+  def set_examples_base(self) -> None:
```
spliffone, line 115: The example url is incorrect - to serve the live-preview examples

pyproject.toml
```diff
@@ -7,9 +7,10 @@ license = "MIT"
 requires-python = ">=3.11, <4"
 readme = "README.md"
 dependencies = [
-    "mkdocs-code-siemens-code-docs-theme>=8.1,<9",
+    "mkdocs-code-siemens-code-docs-theme[zensical]>=8.2.1,<9",
     "mkdocs-minify-html-plugin==0.3.11",
```
spike-rabbit: why do we still need the mkdocs dependencies?

spike-rabbit (changes requested):

pyproject.toml
Killusions: Zensical still supports mkdocs plugins and extensions (just not all hooky for plugins). But minify we could theoretically drop.

Killusions: Dropped.

projects/element-docs-builder/md_extension_element_docs/extension.py
Killusions: Should be fixed now.

Killusions: There was still another bug, now fixed.

mkdocs.yml
```diff
@@ -3,6 +3,7 @@ site_name: Siemens Element Documentation
 site_author: 'Siemens'
 site_url: 'https://element.siemens.io/'
 site_dir: 'dist/design'
+use_directory_urls: !ENV [USE_DIRECTORY_URLS, true]
```
spliffone, line 6: I thought we alway us the directory url approach in github since we have cloudfront on top?

If that is the case we could get rid of some code in the plugin

Killusions: we should probably still pass this here and might need it in the future

Killusions: (basically, we didn't add any new logic for this, just logic for zensical, but previously this option was available via CLI, with zensical not so we expose it in the same way as in other projects).

projects/element-docs-builder/pyproject.toml
```diff
@@ -11,8 +11,12 @@ requires-python = ">=3.11, <4"
 license = "MIT"
 dependencies = [
     "javascript==1!1.2.6",
+    "mkdocs>=1.6.1",
 ]
 
+[project.optional-dependencies]
+zensical = ["zensical>=0.0.50"]
+
```
spike-rabbit, line 19: why do we have both here? I though zensical replaces mkdocs?

spike-rabbit (dismissed):

projects/element-docs-builder/pyproject.toml
Killusions: It does but our docs-theme still has the dependency, this makes the docs-builder compatible with both.

spike-rabbit: But why should it be compatible with both? We just use this internal. I would prefer a clean migration.

Killusions: @spike-rabbit Our compat shims and the legacy mkdocs material theme also still uses mkdocs stuff. The clean migration will be later.

Killusions:
Also FYI, the code QL warning was valid but is now solved, unfortunately it's sticky

projects/element-docs-builder/pyproject.toml
spike-rabbit: For the mkdocs theme I understand that. But why should we have compatibility? We don't need the compatiblity with mkdocs anymore

Killusions: Zensical uses some mkdocs stuff still under to hood for compat.

Killusions:
We also use Zensical for the code.siemens.io docs now BTW

PR #2563, opened 2026-08-17
chore: upgrade to pnpm 12

pwuersch:
Upgrades pnpm to v12 and switches out the combination of `pnpm/action-setup` and `actions/setup-node` with the new `pnpm/setup` action: https://github.com/pnpm/setup#setup-pnpm-with-runtime

### DIFF
--- a/.github/workflows/build-and-test.yaml
+++ b/.github/workflows/build-and-test.yaml
@@ -22,11 +22,12 @@ jobs:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
         with:
           fetch-depth: 0 # We need this for commit-lint
-      - uses: pnpm/action-setup@0977fd99725f1db4007ccb2928dbb4e90d06cc86 # v6.0.10
-      - uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7
+      - name: Install Node.js and pnpm
+        uses: pnpm/setup@84cb39b217b10273981911c288cd62326dc7c6d2 # v2.0.2
         with:
-          node-version: lts/krypton
-          cache: 'pnpm'
+          runtime: node@krypton
+          cache: true
+          install: false
       - name: Set up registry authentication
         run: pnpm config set "//code.siemens.com/:_authToken" "${SIEMENS_NPM_TOKEN}"
         env:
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -13,7 +13,7 @@
     "email": "info@siemens.com"
   },
   "type": "module",
-  "packageManager": "pnpm@11.20.0",
+  "packageManager": "pnpm@12.0.0-rc.6",
```
pwuersch, line 16: I'll revisit it when the stable release comes out. Local testing looks extremely promising and suggests a big performance improvement.

pwuersch: [pnpm v12](https://github.com/pnpm/pnpm/releases/tag/v12.0.0) is the stable release, which I updated the PR to now.

pwuersch:
Hey @spliffone, pinging you here because you initially went on the pnpm tour in #1933. I think the CI workflows can't pass because I'm not really onboarded to the repo and hence don't have access to the tokens that pull from code.siemens.com. Can you give this a try?

.github/workflows/build-and-test.yaml
```diff
@@ -22,11 +22,12 @@ jobs:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
         with:
           fetch-depth: 0 # We need this for commit-lint
-      - uses: pnpm/action-setup@0977fd99725f1db4007ccb2928dbb4e90d06cc86 # v6.0.10
-      - uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7
+      - name: Install Node.js and pnpm
+        uses: pnpm/setup@84cb39b217b10273981911c288cd62326dc7c6d2 # v2.0.2
```
spliffone, line 26: Can we use a tagged version, I have troubles to find the commit hash

pwuersch: This is already a tagged version `v2.0.2`. I got the commit hash from the release: https://github.com/pnpm/setup/releases/tag/v2.0.2. The previous `pnpm/action-setup` also used patch-specific versions; that's why I added it here too. We could also use a major version specifier like this:

```suggestion
        uses: pnpm/setup@84cb39b217b10273981911c288cd62326dc7c6d2 # v2
```

Did I understand your question correctly, @spliffone?

spliffone:
This PR has been taken over in #2658 so the workflow can run from an upstream branch with access to the repository secrets.

The original commit `dd2d5976fc16eb0d321883c0f7aabac79404439e` and @pwuersch's authorship are preserved unchanged. Continuing review and CI in #2658.

PR #2591, opened 2026-08-20
feat(theme): support dynamic dom target in `SiThemeService`

chintankavathia:
Allows `SiThemeService` to target specified DOM root instead of hardcoded document element which fails in case of micro frontends loading element on shadow dom host. `siShadowRoot` can now preconfigure target element for `SiThemeService`.

## Background
The theme service previously applied theme state only to the document root. Content rendered in a Shadow DOM using :host as root selector therefore did not reliably receive theme classes, color scheme, or generated theme styles.

---

[Documentation](https://d33c9dcnqinn2a.cloudfront.net/pr-2591/pages/).
[Examples](https://d33c9dcnqinn2a.cloudfront.net/pr-2591/pages/element-examples/).
[Dashboards Demo](https://d33c9dcnqinn2a.cloudfront.net/pr-2591/pages/dashboards-demo/).
[Playwright report](https://d33c9dcnqinn2a.cloudfront.net/pr-2591/playwright-report/).

**Coverage Reports:**

Code Coverage

- [charts-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-2591/pages/coverage/charts-ng/)
- [dashboards-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-2591/pages/coverage/dashboards-ng/)
- [element-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-2591/pages/coverage/element-ng/)
- [element-translate-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-2591/pages/coverage/element-translate-ng/)
- [maps-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-2591/pages/coverage/maps-ng/)
- [native-charts-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-2591/pages/coverage/native-charts-ng/)


projects/element-ng/theme/si-theme-dom-target.ts
```diff
@@ -0,0 +1,48 @@
+/**
+ * Copyright (c) Siemens 2016 - 2026
+ * SPDX-License-Identifier: MIT
+ */
+import { DOCUMENT, inject, InjectionToken } from '@angular/core';
+import { Meta } from '@angular/platform-browser';
+
+export type SiResolvedColorScheme = 'dark' | 'light';
```
dr-itz: In theme service we have `type ThemeType = 'dark' | 'light' | 'auto'`. So this is the same minus the 'auto'.

Can we instead have something like

```ts
export type ColorScheme = 'light' | 'dark';
export type ThemeType = ColorScheme | 'auto'
```

so that the relation is clear. Guess si-theme.model.ts is the right place for this

chintankavathia: Done as suggested

dr-itz (approved):

projects/element-ng/shadow-root/si-shadow-root.directive.ts
```diff
@@ -33,17 +37,41 @@ import { Directive, ElementRef, inject, DOCUMENT, DestroyRef } from '@angular/co
  */
 @Directive({
   selector: '[siShadowRoot]',
-  providers: [{ provide: OverlayContainer, useExisting: SiShadowRootDirective }, Overlay]
+  providers: [
+    { provide: OverlayContainer, useExisting: SiShadowRootDirective },
+    Overlay,
+    SiThemeService,
+    { provide: SI_THEME_DOM_TARGET, useFactory: createShadowRootThemeTarget }
```
spike-rabbit, line 44: I am not sure, if it is really correct to tie this together with the theme service.

I think a valid use case could be, the an application has the theme / tokens defined on the :root and only loads the element-theme per shadow root.
That way, they can ensure that for all shadow roots, each token has the same value.
I that case `siShadowRoot` should be used, to ensure overlays have the correct element-theme, but it should still use the :root theme.

spike-rabbit (changes requested):
In case an application is actually needing this, shouldn't we somehow allow to sync the dark/light state globally?

projects/element-ng/shadow-root/si-shadow-root.directive.ts
chintankavathia: use case here is that parent application may not have the same version of element as shadow root (mostly separate microfront end app) and in that case it shouldn't rely on tokens defined on :root.

Issue #2619, opened 2026-08-25
CDK overlay uses native popovers by default, breaking z-index based layering


dr-itz:
### Prerequisites

- [x] I have read the [Contributing Guidelines](https://github.com/siemens/element/blob/main/CONTRIBUTING.md).
- [x] I have not leaked any internal/restricted information like screenshots, videos, code snippets, links etc.

### Summary

CDK enabled the use of native popovers with 21.0.6. All our z-index tricks break because of this.

E.g. a toast inside a full-screen modal will not show any more. In our si-tour demo, the menu that opens should not be interactive, but it is. We can somewhat work around this by passing `usePopover: false` to our own uses of overlay, but any consumer created overlays (including cdkConnectedOverlay directive) will not be affected by this.

PR #2622, opened 2026-08-25
fix: bring native popover to top layer for special cases

dr-itz:
See #2619

---

[Documentation](https://d33c9dcnqinn2a.cloudfront.net/pr-2622/pages/).
[Examples](https://d33c9dcnqinn2a.cloudfront.net/pr-2622/pages/element-examples/).
[Dashboards Demo](https://d33c9dcnqinn2a.cloudfront.net/pr-2622/pages/dashboards-demo/).
[Playwright report](https://d33c9dcnqinn2a.cloudfront.net/pr-2622/playwright-report/).

**Coverage Reports:**

Code Coverage

- [charts-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-2622/pages/coverage/charts-ng/)
- [dashboards-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-2622/pages/coverage/dashboards-ng/)
- [element-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-2622/pages/coverage/element-ng/)
- [element-translate-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-2622/pages/coverage/element-translate-ng/)
- [maps-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-2622/pages/coverage/maps-ng/)
- [native-charts-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-2622/pages/coverage/native-charts-ng/)


dr-itz:
@spike-rabbit WDYT?

In case there are toast and a full-screen modal is opened, it will be in front of the toasts until another toast is displayed. I think this behaviour is actually good or at least better than what we have now (not showing the toast at all).

This would also need a v49 backport

spike-rabbit:
I guess for now we don't have any choice.
What about we request a feature for the CDK the brings an alternative for the z-index base stacking.
So the cdk itself brings overlays in the correct order? Something like a priority config or so.

dr-itz:
> I guess for now we don't have any choice. What about we request a feature for the CDK the brings an alternative for the z-index base stacking. So the cdk itself brings overlays in the correct order? Something like a priority config or so.

since the z-index based stacking is our own (disgusting) hack, from a CDK POV there's no change in using popover. Basically stacking is always in order of opening the overlay. And with popover API it's the same. And there is now way to change stacking order with the popover API.

spike-rabbit (approved):
👍

spike-rabbit (changes requested):
Can you add you add a test for this?
e2e is also fine if needed

playwright/e2e/element-examples/si-modal-service.spec.ts
```diff
@@ -28,4 +28,15 @@ test.describe('si-modal-service', () => {
...
     await si.runVisualAndA11yTests('icon');
   });
+
+  test(example + ' full screen modal + toast', async ({ page, si }) => {
+    await si.visitExample(example);
+
+    await page.locator('.btn').getByText('Create full-screen modal').click();
+    await expect(page.locator('.modal-body')).toBeVisible();
+    await page.locator('.btn').getByText('Show manual close toast').click();
+    await expect(page.locator('.toast-content').first()).toBeVisible();
```
spike-rabbit, line 38: please favor using page.getByRole() instead of CSS selectors.
What about taking snapshot while both modal and toast are visible?

spike-rabbit (changes requested):

playwright/e2e/element-examples/si-modal-service.spec.ts
dr-itz: This is all just copied from the same test and the toast test.

Taking the snapshot while to modal is open is exactly what the test does. It's a full-screen modal.

spike-rabbit: ah sorry, I am blind.

spike-rabbit (approved):

PR #2694, opened 2026-09-03
feat(dashboards): add option to hide already added widgets from catalog

chintankavathia:
Add `hideAddedWidgetsFromCatalog` optional input to prevent already-added widget types from appearing in the catalog.

---

[Documentation](https://d33c9dcnqinn2a.cloudfront.net/pr-2694/pages/).
[Examples](https://d33c9dcnqinn2a.cloudfront.net/pr-2694/pages/element-examples/).
[Dashboards Demo](https://d33c9dcnqinn2a.cloudfront.net/pr-2694/pages/dashboards-demo/).
[Playwright report](https://d33c9dcnqinn2a.cloudfront.net/pr-2694/playwright-report/).

**Coverage Reports:**

Code Coverage

- [charts-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-2694/pages/coverage/charts-ng/)
- [dashboards-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-2694/pages/coverage/dashboards-ng/)
- [element-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-2694/pages/coverage/element-ng/)
- [element-translate-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-2694/pages/coverage/element-translate-ng/)
- [maps-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-2694/pages/coverage/maps-ng/)
- [native-charts-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-2694/pages/coverage/native-charts-ng/)

### DIFF
--- a/api-goldens/dashboards-ng/index.api.md
+++ b/api-goldens/dashboards-ng/index.api.md
@@ -197,6 +197,7 @@ export class SiFlexibleDashboardComponent implements OnInit, OnChanges, OnDestro
     readonly editable: _angular_core.ModelSignal<boolean>;
     readonly grid: _angular_core.Signal<SiGridComponent>;
     readonly heading: _angular_core.InputSignal<string | undefined>;
+    readonly hideAddedWidgetsFromCatalog: _angular_core.InputSignalWithTransform<boolean, unknown>;
     readonly hideAddWidgetInstanceButton: _angular_core.InputSignal<boolean>;
     readonly hideEditButton: _angular_core.InputSignal<boolean>;
     readonly hideProgressIndicator: _angular_core.InputSignal<boolean>;
--- a/projects/dashboards-ng/src/components/flexible-dashboard/si-flexible-dashboard.component.spec.ts
+++ b/projects/dashboards-ng/src/components/flexible-dashboard/si-flexible-dashboard.component.spec.ts
@@ -16,9 +16,10 @@ import {
 import { ComponentFixture, TestBed } from '@angular/core/testing';
 import { MenuItem } from '@siemens/element-ng/common';
 import { SiLoadingSpinnerModule } from '@siemens/element-ng/loading-spinner';
-import { firstValueFrom, Observable, of, zip } from 'rxjs';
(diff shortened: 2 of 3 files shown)


timowolf (changes requested):
I am not happy with the workflow. I think most dashboards do not have this requirements, some have it on individual widgets and others on all. I also thing that disabling added widgets might be interesting to still show consistent data but indicate that it is not allowed anymore.

Let us clarify.

panch1739:
> I am not happy with the workflow. I think most dashboards do not have this requirements, some have it on individual widgets and others on all. I also thing that disabling added widgets might be interesting to still show consistent data but indicate that it is not allowed anymore.
> 
> Let us clarify.

@timowolf I think the concern is valid, but I would not solve it by forcing one behavior for the whole catalogue.

Different widgets can have different availability rules. Some may allow multiple instances, some only one, and others may be unavailable because of licensing, permissions, dependencies, or configuration. The catalogue should reflect that current availability rather than treating all unavailable widgets the same way.

For me, the key distinction is between an action that is no longer applicable and an action that is still relevant but currently blocked.

- If a widget allows only one instance and that instance has already been added, the “add” action no longer applies. In that case, I would hide the entry from the catalogue. The widget has not really disappeared from the product; it has moved from the catalogue to the dashboard, where the user can already see it. If the widget is removed again, it becomes available in the catalogue again.
- If a widget can still conceptually be added but something prevents the user from doing so, I would keep it visible and disabled. E.g., “License required”, “Permission required”, or “Configuration missing”. In those cases, showing the widget is useful because it communicates that the capability exists and explains why it cannot currently be used.

The fact that some widgets can be added multiple times and others only once does not feel inconsistent to me. The maximum number of instances is simply a property of each widget. A widget could have a maximum of one, another five, and another unlimited. The catalogue then reacts to that rule.

Available
→ show normally

Multiple instances allowed
→ keep showing while another instance can still be added

Instance limit reached
→ hide from the add catalogue
→ show again if an instance is removed

Potentially available but currently blocked
e.g. license, permission, dependency
→ keep visible
→ disable
→ explain why

PR #2727, opened 2026-09-05
feat(split)!: remove scale input in favour of unit

spliffone:
BREAKING CHANGE: The `scale` input and exported `Scale` type have been removed from `si-split-part`.

Use `unit="fr"` for split parts that scale with the available space and
`unit="px"` for split parts that keep a fixed size.

Before:

```html
<si-split>
  <si-split-part scale="auto">
    Flexible content
  </si-split-part>
  <si-split-part scale="none">
    Fixed content
  </si-split-part>
</si-split>
```

After:

```html
<si-split>
  <si-split-part size="1" unit="fr">
    Flexible content
  </si-split-part>
  <si-split-part size="240" unit="px">
    Fixed content
  </si-split-part>
</si-split>
```

Replace typed Scale values with SplitUnit, mapping 'auto' to 'fr'
and 'none' to 'px'.

---

[Documentation](https://d33c9dcnqinn2a.cloudfront.net/pr-2727/pages/).
[Examples](https://d33c9dcnqinn2a.cloudfront.net/pr-2727/pages/element-examples/).
[Dashboards Demo](https://d33c9dcnqinn2a.cloudfront.net/pr-2727/pages/dashboards-demo/).
[Playwright report](https://d33c9dcnqinn2a.cloudfront.net/pr-2727/playwright-report/).

**Coverage Reports:**

Code Coverage

- [charts-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-2727/pages/coverage/charts-ng/)
- [dashboards-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-2727/pages/coverage/dashboards-ng/)
- [element-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-2727/pages/coverage/element-ng/)
- [element-translate-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-2727/pages/coverage/element-translate-ng/)
- [maps-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-2727/pages/coverage/maps-ng/)
- [native-charts-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-2727/pages/coverage/native-charts-ng/)

### DIFF
--- a/projects/element-ng/schematics/ng-update/index.spec.ts
+++ b/projects/element-ng/schematics/ng-update/index.spec.ts
@@ -754,6 +754,33 @@ export class SplitComponent {
     expect(template).not.toContain("=== 'none'");
   });
 
+  it('should keep dynamic scale mappings valid after their values are migrated', async () => {
+    addTestFiles(appTree, {
+      '/projects/app/src/split.component.ts': `import { Component } from '@angular/core';
+import { Scale } from '@siemens/element-ng/split';
+
+const defaultScale: Scale = 'none';
+
+@Component({
+  selector: 'app-split',
+  template: \`<si-split-part [scale]="assertedScale" />
+<si-split-part [scale]="inheritedScale" />\`
+})
+export class SplitComponent {
+  readonly assertedScale = 'none' as Scale;
+  readonly inheritedScale = defaultScale;
+}`
+    });
+
+    const tree = await runner.runSchematic('migration-v51', {}, appTree);
+    const component = tree.readContent('/projects/app/src/split.component.ts');
+
+    expect(component).toContain("const defaultScale: SplitUnit = 'px';");
+    expect(component).toContain("readonly assertedScale = 'px' as SplitUnit;");
+    expect(component).toContain(`[unit]="['none', 'px'].includes(assertedScale) ? 'px' : 'fr'"`);
+    expect(component).toContain(`[unit]="['none', 'px'].includes(inheritedScale) ? 'px' : 'fr'"`);
+  });
+
   it('should scope inline template bindings to their owning component', async () => {
     addTestFiles(appTree, {
       '/projects/app/src/owners.ts': `import { Component, signal } from '@angular/core';
--- a/projects/element-ng/schematics/ng-update/index.spec.ts
+++ b/projects/element-ng/schematics/ng-update/index.spec.ts
@@ -614,6 +614,46 @@ export class SplitComponent {}`
     expect(component).not.toContain('scale=');
   });
 
+  it('should warn when static scale and unit inputs conflict', async () => {
+    const logSpy = vi.fn();
+    runner.logger.subscribe(logSpy);
+
+    addTestFiles(appTree, {
+      '/projects/app/src/split.component.ts': `import { Component } from '@angular/core';
+
+@Component({
+  selector: 'app-split',
+  template: \`<si-split>
+  <si-split-part unit="fr" scale="none">Fixed conflict</si-split-part>
+  <si-split-part unit="px" scale="auto">Flexible conflict</si-split-part>
+  <si-split-part unit="fr" scale="auto">Matching</si-split-part>
+  <si-split-part [unit]="unit" [scale]="scale">Dynamic</si-split-part>
+</si-split>\`
+})
+export class SplitComponent {
+  readonly unit = 'fr';
+  readonly scale = 'auto';
+}`
+    });
+
+    const tree = await runner.runSchematic('migration-v51', {}, appTree);
+    const component = tree.readContent('/projects/app/src/split.component.ts');
+
+    expect(component).toMatch(/<si-split-part unit="fr"[^>]*>Fixed conflict<\/si-split-part>/);
+    expect(component).toMatch(/<si-split-part unit="px"[^>]*>Flexible conflict<\/si-split-part>/);
+    expect(component).toMatch(/<si-split-part unit="fr"[^>]*>Matching<\/si-split-part>/);
+    expect(component).toMatch(/<si-split-part \[unit\]="unit"[^>]*>Dynamic<\/si-split-part>/);
+    expect(logSpy).toHaveBeenCalledWith(
+      expect.objectContaining({
+        level: 'warn',
+        message: expect.stringContaining('projects/app/src/split.component.ts')
+      })
+    );
+    expect(logSpy).toHaveBeenCalledWith(
+      expect.objectContaining({ message: expect.stringContaining('conflicting scale and unit') })
+    );
+  });
+
   it('should migrate dynamic split scale bindings in external templates', async () => {
     addTestFiles(appTree, {
       '/projects/app/src/split.component.ts': `import { Component } from '@angular/core';
--- a/projects/element-ng/schematics/ng-update/index.spec.ts
+++ b/projects/element-ng/schematics/ng-update/index.spec.ts
@@ -589,4 +589,327 @@ export class SplitComponent {}`
 })
 export class SplitComponent {}`);
   });
+
+  it('should migrate split scale inputs to units in inline templates', async () => {
+    addTestFiles(appTree, {
+      '/projects/app/src/split.component.ts': `import { Component } from '@angular/core';
+
+@Component({
+  selector: 'app-split',
+  template: \`<si-split>
+  <si-split-part scale="auto" size="1">Flexible</si-split-part>
+  <si-split-part scale="none" size="240">Fixed</si-split-part>
+  <si-split-part scale="auto" size="2" unit="px">Explicit unit</si-split-part>
+</si-split>\`
+})
+export class SplitComponent {}`
+    });
+
+    const tree = await runner.runSchematic('migration-v51', {}, appTree);
+    const component = tree.readContent('/projects/app/src/split.component.ts');
+
+    expect(component).toContain('<si-split-part unit="fr" size="1"');
+    expect(component).toContain('<si-split-part unit="px" size="240"');
+    expect(component).toContain('<si-split-part size="2" unit="px"');
+    expect(component).not.toContain('scale=');
+  });
+
+  it('should migrate dynamic split scale bindings in external templates', async () => {
+    addTestFiles(appTree, {
+      '/projects/app/src/split.component.ts': `import { Component } from '@angular/core';
+
+@Component({
+  selector: 'app-split',
+  templateUrl: './split.component.html'
+})
+export class SplitComponent {
+  readonly scale = 'auto';
+  getScale(): string {
+    return this.scale;
+  }
+}`,
+      '/projects/app/src/split.component.html': `<si-split>
+  <si-split-part [scale]="scale" size="1">Flexible</si-split-part>
+  <si-split-part [scale]="getScale()" size="240">Dynamic</si-split-part>
+</si-split>`
+    });
+
+    const tree = await runner.runSchematic('migration-v51', {}, appTree);
+    const template = tree.readContent('/projects/app/src/split.component.html');
+
+    expect(template).toContain(`<si-split-part [unit]="(scale) === 'none' ? 'px' : 'fr'" size="1"`);
+    expect(template).toContain(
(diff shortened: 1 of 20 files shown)


projects/element-ng/schematics/ng-update/migrate-split-scale.ts
```diff
@@ -0,0 +1,429 @@
...
+
+  return formatBoundAttribute('unit', getUnitExpression(attribute.value));
+};
+
+const getScaleUnit = (value: string): 'px' | 'fr' => (value.trim() === 'none' ? 'px' : 'fr');
+
+const getUnitExpression = (expression: string): string => {
+  const value = expression.trim();
+  return `(${value}) === 'none' ? 'px' : 'fr'`;
+};
```
chintankavathia, line 225: Typed dynamic bindings map migrated 'px' back to 'fr'. A Scale = 'none' becomes SplitUnit = 'px', but the template still compares against 'none''

projects/element-ng/schematics/ng-update/migrate-split-scale.ts
```diff
@@ -0,0 +1,429 @@
...
+    }
+
+    if (ts.isPropertyDeclaration(node) && ts.isIdentifier(node.name)) {
+      if (node.type && typeContainsScale(node.type, scaleTypeNames)) {
+        typedNames.add(node.name.text);
+        if (node.initializer) {
+          collectLiterals(node.initializer);
+        }
+      }
+    }
```
chintankavathia: Class fields such as `signal<Scale>('auto')` become `signal<SplitUnit>('auto')`, causing compilation failure. Generic initializer handling only covers variable declarations

projects/element-ng/schematics/ng-update/migrate-split-scale.ts
```diff
@@ -0,0 +1,429 @@
...
+    ts.forEachChild(node, collectLiterals);
+  };
+
+  const visit = (node: ts.Node): void => {
+    if (ts.isVariableDeclaration(node) && ts.isIdentifier(node.name)) {
+      if (node.type && typeContainsScale(node.type, scaleTypeNames)) {
+        typedNames.add(node.name.text);
+        if (node.initializer) {
+          collectLiterals(node.initializer);
```
chintankavathia: Recursive literal collection rewrites unrelated values. For example, `{ scale: 'auto', objectFit: 'none' } ` becomes `{ scale: 'fr', objectFit: 'px' }` when the containing type references Scale

projects/element-ng/schematics/ng-update/migrate-split-scale.ts
```diff
@@ -0,0 +1,429 @@
...
+    if (ts.isIdentifier(node)) {
+      const replacement = replacements.get(node.text);
+      if (replacement) {
+        recorder.remove(node.getStart(sourceFile), node.getWidth(sourceFile));
+        recorder.insertLeft(node.getStart(sourceFile), replacement);
+        return;
+      }
+    }
+
+    ts.forEachChild(node, visit);
```
chintankavathia, line 450: Every identifier named `Scale` is renamed, including unrelated property names. interface Config { Scale: string; split: Scale } becomes { SplitUnit: string; split: SplitUnit }

chintankavathia (changes requested):

chintankavathia (approved):
@kfenner Just check commit message scope as I am not sure if we decided to go with ! format otherwise LGTM

projects/element-ng/schematics/ng-update/migrate-split-scale.ts
```diff
@@ -0,0 +1,653 @@
...
+  }
+
+  return formatUnitBinding(attribute.value, scaleMemberNames, expression);
+};
+
+const getScaleUnit = (value: string): 'px' | 'fr' => (value.trim() === 'none' ? 'px' : 'fr');
+
+const getUnitExpression = (expression: string): string => {
+  const value = expression.trim();
+  return `(${value}) === 'none' ? 'px' : 'fr'`;
```
kfenner: Following up on the earlier comment about typed dynamic bindings — I think this is still reachable.

Literal migration is *type-checker driven* (`collectContextualLiterals` walks everything with a `Scale` contextual type), but member detection is *syntax driven* (`collectScaleMemberNames` only matches an explicit `: Scale` annotation or `signal<Scale>()`). Anything between the two falls through:

```ts
protected scale = 'none' as Scale;   // literal migrated to 'px', member NOT collected
```

The template then gets `[unit]="(scale) === 'none' ? 'px' : 'fr'"`, which evaluates to `'fr'` for the already-migrated `'px'` value. Same happens for a field initialised from a migrated module-level `const`.

You already wrote the idempotent form for the multi-owner fallback in `migrateScaleTemplate`. Reusing it here makes this safe regardless of whether the member was detected:

```suggestion
  return `['none', 'px'].includes(${value}) ? 'px' : 'fr'`;
```

projects/element-ng/schematics/ng-update/migrate-split-scale.ts
```diff
@@ -0,0 +1,653 @@
...
+  recorder: UpdateRecorder,
+  owners: Set<string>[]
+): void => {
+  findElement(template, element => element.name === 'si-split-part').forEach(element => {
+    const scale = element.attrs.find(attribute => scaleAttributeNames.includes(attribute.name));
+    if (!scale) {
+      return;
+    }
+
+    const hasUnit = element.attrs.some(attribute => unitAttributeNames.includes(attribute.name));
```
kfenner: When a `unit` attribute is already present, `scale` is removed without looking at its value and without any warning. Both of these lose behaviour silently:

- `unit="fr" scale="none"` — was fixed-size, becomes growing
- `unit="px" scale="auto"` — was growing, becomes fixed-size

At minimum, `context.logger.warn` when the dropped `scale` contradicts the existing `unit`, so consumers know where to look.

projects/element-ng/schematics/ng-update/migrate-split-scale.ts
```diff
@@ -0,0 +1,653 @@
...
+      removeAttribute(template, scale, offset, recorder);
+      return;
+    }
+
+    const replacements = new Set(owners.map(names => getUnitAttribute(scale, names)));
+    let replacement = replacements.values().next().value!;
+    if (replacements.size > 1) {
+      const expression =
+        scale.name === 'scale'
+          ? /^\s*{{([\s\S]*)}}\s*$/.exec(scale.value)![1]!.trim()
```
kfenner: Two non-null assertions on an unguarded `exec`. As far as I can tell this is currently unreachable (a static, non-interpolated `scale` always collapses to a single owner-independent replacement, so `replacements.size === 1`), but it is a latent crash if that invariant ever shifts. Worth guarding, or capturing the interpolation once in `getUnitAttribute` and passing it down.

projects/element-ng/schematics/ng-update/migrate-split-scale.ts
```diff
@@ -0,0 +1,653 @@
...
+
+  collect(initializer);
+  if (!replacements.length) {
+    return undefined;
+  }
+
+  return replacements
+    .sort((first, second) => second.start - first.start)
+    .reduce((value, replacement) => {
+      const offset = initializer.getStart(source) - expression.indexOf(expression.trim());
```
kfenner: This offset math is correct but very hard to follow — `expression.indexOf(expression.trim())` is effectively the leading-whitespace length of the synthetic `const value = ...` source. A named constant or a short comment explaining what is being subtracted would help a lot here.

projects/element-ng/schematics/ng-update/index.spec.ts
```diff
@@ -589,4 +589,327 @@ export class SplitComponent {}`
...
+    const template = tree.readContent('/projects/app/src/split.component.html');
+
+    expect(template).toContain(`<si-split-part [unit]="(scale) === 'none' ? 'px' : 'fr'" size="1"`);
+    expect(template).toContain(
+      `<si-split-part [unit]="(getScale()) === 'none' ? 'px' : 'fr'" size="240"`
+    );
+    expect(template).not.toContain('[scale]');
+  });
+
+  it('should preserve scale-derived units when migrating split sizes', async () => {
```
kfenner: This test locks in the size-semantics problem from the summary: the `scale="none"` part comes out as `unit="px" size="20"`, where `20` originally meant 20% of the container and now means 20 pixels. Worth revisiting together with the migration behaviour itself.

projects/element-ng/split/si-split.component.ts
```diff
@@ -254,17 +253,22 @@ export class SiSplitComponent {
...
-            );
+            this.emitSizesChange();
           },
           complete: () => this.saveUIState()
         });
     });
   }
 
+  private emitSizesChange(): void {
+    const containerSize = this.measureContainerSize();
```
kfenner, line 264: This turns a once-per-drag measurement into a per-frame one. Previously `containerSize` was hoisted before the drag subscription; now `measureContainerSize()` (→ `getBoundingClientRect()`) runs on every `pointermove`, immediately after `gridTemplateColumns` was written — a forced synchronous layout each frame.

Suggest keeping the hoisted measurement for the drag path and using `emitSizesChange()` only for the restore path (where the `> 0` guard is genuinely useful), or passing the cached size in as a parameter.

projects/element-ng/split/si-split.component.ts
```diff
@@ -318,24 +322,29 @@ export class SiSplitComponent {
...
+      let restored = false;
       this.parts()
         .filter(part => part.stateId())
         .map(part => ({ part, state: uiState[part.stateId()!] }))
         .filter(
           (item): item is { part: SiSplitPartComponent; state: SplitPartState } =>
-            !!item.state &&
-            item.state.initialSize === item.part.size() &&
-            item.state.initialUnit === item.part.unit()
+            !!item.state && item.state.initialUnit === item.part.unit()
```
kfenner, line 331: With `initialSize` dropped from the filter it is still written in `saveUIState` but never read again — dead field in the persisted schema. Either remove it from `SplitPartState` or keep comparing it.

Also worth calling out explicitly in the docs: changing a part's configured `size` in code no longer invalidates a user's stored layout. I assume that is the intent given the commit message, but it is a user-visible behaviour change.

kfenner (commented):
Reviewed the schematic end-to-end. The three earlier review comments about class-field generics, recursive literal collection and identifier renaming are properly fixed and each has a named regression test — nice work. A few things still open:

**Blocking (in my opinion): `scale="none"` silently reinterprets `size` from a fraction to pixels.**

`splitScaleMigrationRule` runs before `splitSizesMigrationRule` in the v51 chain, so:

```html
<si-split [sizes]="[20, 80]">
  <si-split-part scale="none">   -->   <si-split-part unit="px" size="20">
```

`[sizes]` were *relative* (20% of the container). The migrated output is **20 physical pixels**. This is currently asserted as the expected result in `index.spec.ts`.

The PR itself shows the correct value cannot be derived statically — `src/app/examples/si-split/si-split-mixed.html` had to be hand-edited from `size="20" unit="fr"` to `size="200" unit="px"` and `size="40" unit="fr"` to `size="100" unit="px"`.

Suggestion: keep `unit="fr"` for `scale="none"`, emit a `context.logger.warn` listing affected files (and/or insert a `<!-- TODO -->`), and document the manual conversion step in the migration docs. A loud manual step is much better than silent layout breakage in consumer apps.

**Re: the commit-message question** — `feat(split)!:` is correct. `tools/semantic-release/commit-config.js` defines `breakingHeaderPattern: /^(\w*)(?:\((.*)\))?!: (.*)$/`, there is already precedent on `main` (`feat(split)!: migrate showCollapseButton to collapsible`), and the commit also carries a proper `BREAKING CHANGE:` footer, so the major bump and changelog entry work either way.

Remaining inline comments below.

projects/element-ng/split/si-split.component.ts
chintankavathia: We should do this in follow up

projects/element-ng/split/si-split.component.ts
chintankavathia: we do it in follow up

kfenner (approved):
LGTM 👍

PR #2728, opened 2026-09-05
feat(main-detail-container)!: support configurable split unit for main part

spliffone:
Add `mainContainerWidthUnit` input to configure the unit of the underlying main
split part when `resizableParts` is enabled. The detail split part always uses
`unit="fr"`.

BREAKING CHANGE: The default sizing unit of the main split part in
`si-main-detail-container` when `resizableParts` is enabled has changed from `fr`
to `px`.

When `resizableParts` is enabled:
- The main part now defaults to `unit="px"` (initial size of `minMainSize`,
  default 300px), and the detail part takes the remaining space (`1fr`).
- `mainContainerWidth` is now interpreted in `mainContainerWidthUnit`. With the
  default `mainContainerWidthUnit="px"`, numeric values are treated as pixels
  instead of percentages.
- `[(mainContainerWidth)]` and `(mainContainerWidthChange)` now emit and accept
  pixel values instead of percentage values.
- When adopting the new `px` default, main-part sizes saved via `stateId` with
  `fr` are not restored because the persisted unit no longer matches. Keeping
  `mainContainerWidthUnit="fr"` preserves compatible saved splitter positions.

To retain the previous relative percentage-based split behavior, explicitly set
`mainContainerWidthUnit="fr"`.

Static layouts (`resizableParts="false"`) are unaffected and continue to
interpret `mainContainerWidth` as a percentage.

Before:

```html
<!-- Main part was 40% width, detail was 60% -->
<si-main-detail-container
  resizableParts
  [mainContainerWidth]="40"
  [(detailsActive)]="detailsActive"
>
  <div slot="mainData">...</div>
  <div slot="details">...</div>
</si-main-detail-container>
```

After (retaining previous relative behavior):

```html
<!-- Set mainContainerWidthUnit="fr" to retain fractional/percentage sizing -->
<si-main-detail-container
  resizableParts
  mainContainerWidthUnit="fr"
  [mainContainerWidth]="40"
  [(detailsActive)]="detailsActive"
>
  <div slot="mainData">...</div>
  <div slot="details">...</div>
</si-main-detail-container>
```

After (adopting new fixed-width layout):
```html
<!-- Main part is fixed pixel width (e.g. 360px), detail expands with remaining space -->
<si-main-detail-container
  resizableParts
  [mainContainerWidth]="360"
  [(detailsActive)]="detailsActive"
>
  <div slot="mainData">...</div>
  <div slot="details">...</div>
</si-main-detail-container>
```

---

[Documentation](https://d33c9dcnqinn2a.cloudfront.net/pr-2728/pages/).
[Examples](https://d33c9dcnqinn2a.cloudfront.net/pr-2728/pages/element-examples/).
[Dashboards Demo](https://d33c9dcnqinn2a.cloudfront.net/pr-2728/pages/dashboards-demo/).
[Playwright report](https://d33c9dcnqinn2a.cloudfront.net/pr-2728/playwright-report/).

**Coverage Reports:**

Code Coverage

- [charts-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-2728/pages/coverage/charts-ng/)
- [dashboards-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-2728/pages/coverage/dashboards-ng/)
- [element-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-2728/pages/coverage/element-ng/)
- [element-translate-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-2728/pages/coverage/element-translate-ng/)
- [maps-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-2728/pages/coverage/maps-ng/)
- [native-charts-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-2728/pages/coverage/native-charts-ng/)

### DIFF
--- a/api-goldens/element-ng/main-detail-container/index.api.md
+++ b/api-goldens/element-ng/main-detail-container/index.api.md
@@ -24,6 +24,7 @@ export class SiMainDetailContainerComponent implements OnInit, OnChanges {
     readonly largeLayoutBreakpoint: _angular_core.InputSignal<number>;
     readonly mainContainerClass: _angular_core.InputSignal<string>;
     readonly mainContainerWidth: _angular_core.ModelSignal<number | "default">;
+    readonly mainContainerWidthUnit: _angular_core.InputSignal<SplitUnit>;
     readonly minDetailSize: _angular_core.InputSignal<number>;
     readonly minMainSize: _angular_core.InputSignal<number>;
     readonly resizableParts: _angular_core.InputSignalWithTransform<boolean, unknown>;
--- a/projects/element-ng/main-detail-container/si-main-detail-container.component.spec.ts
+++ b/projects/element-ng/main-detail-container/si-main-detail-container.component.spec.ts
@@ -157,6 +157,7 @@ describe('MainDetailContainerComponent', () => {
   });
 
   it('should update mainContainerWidth in pixels when split sizes change with px unit', async () => {
+    vi.useRealTimers();
     const widthSpy = vi.spyOn(component, 'mainContainerWidthChanged');
     component.resizableParts.set(true);
     await fixture.whenStable();
@@ -166,12 +167,13 @@ describe('MainDetailContainerComponent', () => {
 
     const split = debugElement.query(By.directive(SiSplitComponent));
     split.componentInstance.sizesChange.emit([38.5, 61.5]);
-    fixture.detectChanges();
+    await fixture.whenStable();
 
     expect(widthSpy).toHaveBeenCalledWith(385);
   });
 
   it('should calculate pixel size from split width when mainSplitPart expandedSize is unavailable', async () => {
+    vi.useRealTimers();
     const widthSpy = vi.spyOn(component, 'mainContainerWidthChanged');
     component.resizableParts.set(true);
     await fixture.whenStable();
@@ -182,12 +184,13 @@ describe('MainDetailContainerComponent', () => {
     const split = debugElement.query(By.directive(SiSplitComponent));
     const splitWidth = split.nativeElement.getBoundingClientRect().width;
     split.componentInstance.sizesChange.emit([30, 70]);
-    fixture.detectChanges();
+    await fixture.whenStable();
 
     expect(widthSpy).toHaveBeenCalledWith(Math.round((splitWidth * 30) / 100));
   });
 
   it('should fallback to minMainSize when split width is 0 and expandedSize is unavailable', async () => {
+    vi.useRealTimers();
     const widthSpy = vi.spyOn(component, 'mainContainerWidthChanged');
     component.resizableParts.set(true);
     await fixture.whenStable();
@@ -198,12 +201,13 @@ describe('MainDetailContainerComponent', () => {
     const split = debugElement.query(By.directive(SiSplitComponent));
     vi.spyOn(split.nativeElement, 'getBoundingClientRect').mockReturnValue({ width: 0 } as DOMRect);
     split.componentInstance.sizesChange.emit([30, 70]);
-    fixture.detectChanges();
+    await fixture.whenStable();
 
     expect(widthSpy).toHaveBeenCalledWith(300);
   });
 
   it('should update mainContainerWidth as percentage when split sizes change with fr unit', async () => {
+    vi.useRealTimers();
     const widthSpy = vi.spyOn(component, 'mainContainerWidthChanged');
     component.mainContainerWidthUnit.set('fr');
     component.resizableParts.set(true);
@@ -211,12 +215,12 @@ describe('MainDetailContainerComponent', () => {
 
     const split = debugElement.query(By.directive(SiSplitComponent));
     split.componentInstance.sizesChange.emit([42, 58]);
-    fixture.detectChanges();
+    await fixture.whenStable();
 
     expect(widthSpy).toHaveBeenCalledWith(42);
   });
 
-  it('should keep the standard static layout for a px-sized main part', async () => {
+  it('should not apply max sizes when mainContainerWidth is outside 0-100', async () => {
     component.mainContainerWidth.set(320);
     await fixture.whenStable();
 
--- a/projects/element-ng/main-detail-container/si-main-detail-container.component.spec.ts
+++ b/projects/element-ng/main-detail-container/si-main-detail-container.component.spec.ts
@@ -13,6 +13,7 @@ import {
   ElementDimensions,
   ResizeObserverService
 } from '../resize-observer';
+import { SiSplitComponent, SiSplitPartComponent, SplitUnit } from '../split';
 import { SiMainDetailContainerComponent } from './si-main-detail-container.component';
 
 @Component({
@@ -27,6 +28,8 @@ import { SiMainDetailContainerComponent } from './si-main-detail-container.compo
       [hideBackButton]="hideBackButton()"
       [largeLayoutBreakpoint]="largeLayoutBreakpoint"
       [resizableParts]="resizableParts()"
+      [mainContainerWidthUnit]="mainContainerWidthUnit()"
+      [mainContainerWidth]="mainContainerWidth()"
       [(detailsActive)]="detailsActive"
       (hasLargeSizeChange)="hasLargeSizeChanged($event)"
       (mainContainerWidthChange)="mainContainerWidthChanged($event)"
@@ -47,6 +50,8 @@ class WrapperComponent {
   readonly detailsHeading = signal('details-heading');
   readonly hideBackButton = signal(false);
   readonly resizableParts = signal(false);
+  readonly mainContainerWidthUnit = signal<SplitUnit>('px');
+  readonly mainContainerWidth = signal<number | 'default'>('default');
   largeLayoutBreakpoint = BOOTSTRAP_BREAKPOINTS.mdMinimum;
   readonly detailsActive = signal(false);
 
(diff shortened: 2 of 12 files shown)


projects/element-ng/main-detail-container/si-main-detail-container.component.ts
```diff
@@ -162,10 +165,19 @@ export class SiMainDetailContainerComponent implements OnInit, OnChanges {
...
 
   /**
-   * The percentage width of the main container from the overall component width.
-   * Can be a number or `'default'`, which is 32% when {@link resizableParts} is active, otherwise 50%.
+   * Unit used for the main split part when {@link resizableParts} is enabled.
    *
-   * @defaultValue 'default'
+   * @defaultValue 'px'
+   */
+  readonly mainUnit = input<SplitUnit>('px');
```
chintankavathia: I am little bit unsure of naming this as `mainUnit`. I would go with something like `mainContainerWidthUnit`. WDYT @kfenner

kfenner: @chintankavathia Fully agree, please go with the new name 🙏

chintankavathia: done

chintankavathia (approved):
👍

projects/element-ng/schematics/ng-update/migrate-main-detail-units.ts
```diff
@@ -0,0 +1,188 @@
...
+  getTemplateUrl
+} from '../utils/index.js';
+
+const resizableAttributeNames = ['resizableParts', '[resizableParts]', 'bind-resizableParts'];
+const mainUnitAttributeNames = [
+  'mainContainerWidthUnit',
+  '[mainContainerWidthUnit]',
+  'bind-mainContainerWidthUnit'
+];
+const detailUnitAttributeNames = ['detailUnit', '[detailUnit]', 'bind-detailUnit'];
```
kfenner: `detailUnit` does not exist anywhere in the codebase — not on this branch, not on the base, not in the public API or the goldens. I grepped both refs to be sure.

So the removal branch below (lines 80-85) is dead code that would silently delete a consumer attribute if it ever matched, and nothing in the spec covers it. Looks like a leftover from an earlier design iteration where the detail part was configurable too.

Could we drop `detailUnitAttributeNames` and the corresponding `removeAttribute` call? If it is intentionally targeting something I have missed, a comment explaining what would help.

projects/element-ng/schematics/ng-update/migrate-main-detail-units.ts
```diff
@@ -0,0 +1,188 @@
...
+
+  const hasMainUnit = element.attrs.some(attr => mainUnitAttributeNames.includes(attr.name));
+  if (hasMainUnit) {
+    return;
+  }
+
+  insertAttribute(template, element, 'mainContainerWidthUnit="fr"', offset, recorder);
+};
+
+const insertAttribute = (
```
kfenner: `insertAttribute` and `removeAttribute` are verbatim copies of the ones already exported from `schematics/utils/template-edit-utils.ts` — and this file already imports `discoverSourceFiles`, `findElement`, `getInlineTemplates` and `getTemplateUrl` from `../utils/index.js`, so they are one import away:

```ts
import { discoverSourceFiles, findElement, getInlineTemplates, getTemplateUrl, insertAttribute, removeAttribute } from '../utils/index.js';
```

While we are here: `getExpression` (line 176) is now duplicated a **third** time — it also lives in `migrate-split-collapse.ts:219` and `migrate-split-scale.ts:640` — and `getStaticBooleanValue` (line 145) a second time (`migrate-split-collapse.ts:170`). Both feel like they belong in `schematics/utils/` alongside the attribute helpers. Happy for that to be a separate cleanup PR if you would rather keep this one focused.

projects/element-ng/main-detail-container/si-main-detail-container.component.ts
```diff
@@ -229,8 +242,14 @@ export class SiMainDetailContainerComponent implements OnInit, OnChanges {
...
       this.doAnimation(changes.detailsActive.currentValue);
     }
-    if (changes.mainContainerWidth || changes.resizableParts) {
-      this.splitSizes = [this.actualMainContainerWidth(), 100 - this.actualMainContainerWidth()];
+    if (
+      changes.mainContainerWidth ||
+      changes.resizableParts ||
+      changes.mainContainerWidthUnit ||
+      changes.minMainSize ||
+      changes.minDetailSize
```
kfenner: `minDetailSize` does not influence `getSplitSizes()` or `getMaxSize()`, so adding it here triggers a recomputation that can never change anything. Harmless, but it implies a dependency that does not exist.

(`minMainSize` and `mainContainerWidthUnit` on the other hand are both genuinely needed — easy to miss, good catch adding them.)

projects/element-ng/main-detail-container/si-main-detail-container.component.ts
```diff
@@ -256,12 +279,42 @@ export class SiMainDetailContainerComponent implements OnInit, OnChanges {
...
+    const mainSize = this.actualMainContainerWidth();
+    if (!this.resizableParts()) {
+      return mainSize >= 0 && mainSize <= 100 ? [mainSize, 100 - mainSize] : [50, 50];
+    }
+
+    if (this.mainContainerWidthUnit() === 'fr') {
+      return mainSize >= 0 && mainSize <= 100 ? [mainSize, 100 - mainSize] : [32, 68];
+    }
+
+    return [mainSize, 1];
```
kfenner, line 314: Two things about this method:

**1. The returned tuple now mixes units.** `[mainSize, 1]` is `[px, fr]`, whereas both branches above return `[percentage, percentage]`. It is valid — `si-split` resolves `unit` per part in `gridTemplate`, so the detail part correctly becomes `1fr` — but the `[number, number]` type carries no hint of that. A short comment on this return, or a `{ main, detail }` shape, would make it much harder to misread later.

**2. The out-of-range fallbacks are silent.** `[50, 50]` and `[32, 68]` swallow invalid input with no dev-mode warning. Worth noting that `fr` weights are legitimately arbitrary in CSS, so someone writing `[mainContainerWidth]="1"` expecting `1fr` silently gets `[1, 99]`. If 0-100 is a hard constraint for both units, it should be explicit in the `mainContainerWidth` JSDoc.

projects/element-ng/main-detail-container/si-main-detail-container.component.ts
```diff
@@ -256,12 +279,42 @@ export class SiMainDetailContainerComponent implements OnInit, OnChanges {
...
+  }
+
+  private getMainSizePx(fallbackPercentage: number): number {
+    const mainPartSize = this.mainSplitPart()?.expandedSize();
+    if (mainPartSize !== undefined) {
+      return Math.round(mainPartSize);
+    }
+
+    const splitWidth = this.splitComponent()?.nativeElement.getBoundingClientRect().width;
```
kfenner: I think both fallbacks here are unreachable in production.

`SiSplitPartComponent.expandedSize` is a `linkedSignal` whose computation is `source.unit === 'fr' ? undefined : source.size` — so whenever the unit is `px` it is **always** a number. And `getMainSizePx` is only called from `onSplitSizesChange` when `mainContainerWidthUnit() === 'px'`, which in turn only fires from a rendered `si-split`, so `mainSplitPart()` is always resolved by then.

That means `mainPartSize !== undefined` is always true, and neither the `getBoundingClientRect` branch nor the `minMainSize()` branch can run. Consistent with that, the two specs covering them only get there by calling `parts[0].componentInstance.expandedSize.set(undefined)` — forcing an `@internal` signal into a state the component itself cannot produce.

Suggest either collapsing this to `Math.round(this.mainSplitPart()!.expandedSize()!)` and dropping those two tests, or adding a comment describing the timing window this is guarding against so the tests read as intentional.

projects/element-ng/main-detail-container/si-main-detail-container.component.spec.ts
```diff
@@ -127,6 +132,98 @@ describe('MainDetailContainerComponent', () => {
...
+    expect(widthSpy).toHaveBeenCalledWith(385);
+  });
+
+  it('should calculate pixel size from split width when mainSplitPart expandedSize is unavailable', async () => {
+    const widthSpy = vi.spyOn(component, 'mainContainerWidthChanged');
+    component.resizableParts.set(true);
+    await fixture.whenStable();
+
+    const parts = debugElement.queryAll(By.directive(SiSplitPartComponent));
+    parts[0]!.componentInstance.expandedSize.set(undefined);
```
kfenner, line 182: This and the following test drive an `@internal` signal into a state the component cannot reach on its own (see my comment on `getMainSizePx`). They will keep passing even if the production fallback is removed, so they are not really protecting anything.

If the fallback stays, it would be more valuable to test it through a reachable path; if it goes, these two can go with it.

projects/element-ng/main-detail-container/si-main-detail-container.component.spec.ts
```diff
@@ -127,6 +132,98 @@ describe('MainDetailContainerComponent', () => {
...
+    await fixture.whenStable();
+
+    const split = debugElement.query(By.directive(SiSplitComponent));
+    split.componentInstance.sizesChange.emit([42, 58]);
+    fixture.detectChanges();
+
+    expect(widthSpy).toHaveBeenCalledWith(42);
+  });
+
+  it('should keep the standard static layout for a px-sized main part', async () => {
```
kfenner: The name does not quite match what is asserted — `resizableParts` is left at `false` here, so there is no "px-sized main part" involved. What this actually verifies is that `mainContainerWidth` outside 0-100 no longer produces a max size in the static layout, i.e. the new guard in `getMaxSize`.

Something like `should not apply max sizes when mainContainerWidth is outside 0-100` would describe it better. Worth keeping either way — it covers the `-220%` regression nicely.

Minor, and applies to the other new tests too: they use `fixture.detectChanges()` after `sizesChange.emit(...)`, while the rest of this file uses `await fixture.whenStable()`.

api-goldens/element-ng/main-detail-container/index.api.md
```diff
@@ -24,6 +24,7 @@ export class SiMainDetailContainerComponent implements OnInit, OnChanges {
     readonly largeLayoutBreakpoint: _angular_core.InputSignal<number>;
     readonly mainContainerClass: _angular_core.InputSignal<string>;
     readonly mainContainerWidth: _angular_core.ModelSignal<number | "default">;
+    readonly mainContainerWidthUnit: _angular_core.InputSignal<SplitUnit>;
```
kfenner, line 27: This golden now references `SplitUnit` but has no corresponding import line, so the type is dangling.

Other goldens do emit cross-entry-point imports when the type resolves — `popover-next` for instance imports `OverlayArrowPosition` from `@siemens/element-ng/common`. That difference suggests api-extractor could not resolve `SplitUnit` from this entry point and may be emitting an `ae-forgotten-export` warning.

Worth checking the extractor log. If it is a genuine forgotten export, `SplitUnit` probably needs re-exporting from the `main-detail-container` entry point, otherwise consumers cannot type a variable they pass into this input.

kfenner (commented):
Reviewed the delta against its stack base (`stack/split-remove-scale-1`).

Overall the design is sound: `px` main + `1fr` detail is the right default for a list/detail layout, `mainContainerWidthUnit="fr"` is a well-chosen escape hatch, and shipping an ng-update migration that injects `fr` is exactly the right compatibility strategy. The migration itself is nicely conservative — treating a dynamic `[resizableParts]="x"` binding as "possibly enabled" and adding the attribute anyway is the correct call, and `getStaticBooleanValue` handles `""`, `"false"`, `[x]="null"` and interpolation properly.

I took the liberty of fixing the PR description directly: it consistently said `mainUnit`, but the implemented input is `mainContainerWidthUnit`. Since the BREAKING CHANGE block and the "retain previous behavior" snippet both end up in the changelog, that would have pointed consumers at a non-existent input. Please double-check my edit. The same stale name is still in the commit message body:

```html
<!-- Set mainUnit="fr" to retain fractional/percentage sizing -->
```

I also added a bullet to the BREAKING CHANGE block about persisted state: `SiSplitComponent.restoreFromUIState()` discards state when `initialUnit !== part.unit()`. Every consumer using `stateId` together with `resizableParts` has `'fr'` persisted today, so all saved splitter positions get dropped on the first load after upgrading. That is the right behavior, but it is user-visible and worth calling out.

Remaining points are inline. The two I would most like to see addressed are the dead `detailUnit` handling and the duplicated schematic helpers.

### Not blocking, but worth a thought

**No example or VRT covers the new default.** Both examples were updated to `mainContainerWidthUnit="fr"`, so nothing in the demo app, docs or e2e actually exercises the new `px` default that this breaking change introduces. Could one of them adopt `px` instead? Relatedly, both examples still log `` `Main width is ${containerWidth}%.` ``, which is only correct because they opt out.

**Emission precision differs by unit.** `px` emits `Math.round(...)`, `fr` emits raw floats (`38.5123...`). Consider rounding both for consistency.

**`ngOnChanges` recomputation keeps growing.** `splitSizes` / `maxMainSize` / `maxDetailSize` are still recomputed via a manually maintained input list, and this PR adds three more entries to it. Everything involved except `hasLargeSize` is already a signal — converting that one field would let all three become `computed()` and delete the block entirely. Good follow-up rather than something for this PR.

**Nits:** `viewChild('splitComponent', { read: ElementRef })` reads an `ElementRef`, not a component — `#splitEl` would read better. And the `@defaultValue 'default', which uses {@link minMainSize} for \`px\` and \`32\` for \`fr\`.` prose tail is unusual for this repo; worth confirming the docs generator renders it.

### Nice catch

The new `mainContainerWidth < 0 || > 100` guard in `getMaxSize` fixes a real latent bug — previously a value like `320` produced `max-inline-size: -220%` on the detail container.

projects/element-ng/schematics/ng-update/migrate-main-detail-units.ts
chintankavathia: fixed

projects/element-ng/schematics/ng-update/migrate-main-detail-units.ts
chintankavathia: reused `insertAttribute`. `removeAttribute` is removed. parser will require update in other schematics so not changing for now.

projects/element-ng/main-detail-container/si-main-detail-container.component.ts
chintankavathia: fixed

projects/element-ng/main-detail-container/si-main-detail-container.component.ts
chintankavathia: updated jsdoc

projects/element-ng/main-detail-container/si-main-detail-container.component.ts
chintankavathia: keeping it for now as they don't harm and are just defensive code

projects/element-ng/main-detail-container/si-main-detail-container.component.spec.ts
chintankavathia: updated

api-goldens/element-ng/main-detail-container/index.api.md
chintankavathia: this is not valid.

projects/element-ng/main-detail-container/si-main-detail-container.component.spec.ts
chintankavathia: not changing as per above comment

kfenner (approved):
LGTM 👍

kfenner (approved):

timowolf (approved):

PR #2729, opened 2026-09-05
feat(list-details)!: support configurable split unit for list pane

spliffone:
Add `listWidthUnit` input to configure the unit of the underlying list split part
when resizing is enabled. The details split part always uses `unit="fr"`.

BREAKING CHANGE: The default sizing unit of the list split part in
`si-list-details` when resizing is enabled has changed from `fr` to `px`.

When resizing is enabled:
- The list part now defaults to `unit="px"` (initial size of 300px), and the
  details part takes the remaining space (`1fr`).
- `listWidth` is now interpreted in `listWidthUnit`. With the default
  `listWidthUnit="px"`, numeric values are treated as pixels instead of percentages.
- `[(listWidth)]` now emits and accepts pixel values instead of percentage
  values.

To retain the previous relative percentage-based split behavior, explicitly set
`listUnit="fr"`.

Static layouts (`disableResizing="true"`) continue to interpret `listWidth` as a
percentage. Values in the inclusive range 0-100 and the default 32/68 layout
are unchanged. Out-of-range values now fall back to a 32/68 list/details split
instead of being passed directly to CSS. This fallback also applies when
resizing is enabled with `listUnit="fr"`.

Before:

```html
<si-list-details [listWidth]="50" />
```

After (retaining previous relative behavior):

```html
<si-list-details
  [listWidth]="50"
  listWidthUnit="fr"
/>
```

After (adopting new fixed-width layout):

```html
<si-list-details [listWidth]="300" />
```

---

[Documentation](https://d33c9dcnqinn2a.cloudfront.net/pr-2729/pages/).
[Examples](https://d33c9dcnqinn2a.cloudfront.net/pr-2729/pages/element-examples/).
[Dashboards Demo](https://d33c9dcnqinn2a.cloudfront.net/pr-2729/pages/dashboards-demo/).
[Playwright report](https://d33c9dcnqinn2a.cloudfront.net/pr-2729/playwright-report/).

**Coverage Reports:**

Code Coverage

- [charts-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-2729/pages/coverage/charts-ng/)
- [dashboards-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-2729/pages/coverage/dashboards-ng/)
- [element-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-2729/pages/coverage/element-ng/)
- [element-translate-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-2729/pages/coverage/element-translate-ng/)
- [maps-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-2729/pages/coverage/maps-ng/)
- [native-charts-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-2729/pages/coverage/native-charts-ng/)

### DIFF
--- a/docs/components/layout-navigation/list-details.md
+++ b/docs/components/layout-navigation/list-details.md
@@ -148,6 +148,16 @@ Inside the `<si-details-pane>` the following child components can be used:
 
 It is highly advised to use at least the `<si-details-pane-header>` to show the back button in a responsive (mobile) view.
 
+When resizing is enabled, the underlying list split part uses the `listUnit`
+input, while the details split part always uses `unit="fr"`. The list pane uses
+`unit="px"` by default. Accordingly, `listWidth` defaults to `300` and
+represents pixels with the default configuration.
+
+Use `listUnit="fr"` to retain a relative split, where `listWidth` is the list
+pane's fractional weight. When resizing is disabled, `listWidth` continues to
+control the static layout as a percentage. Values outside the percentage range
+use the standard 32/68 static layout.
+
 If the content may exceed the available space, the `overflow-auto` class should be applied
 to the body components or a child inside them.
 
--- a/docs/components/layout-navigation/list-details.md
+++ b/docs/components/layout-navigation/list-details.md
@@ -148,14 +148,14 @@ Inside the `<si-details-pane>` the following child components can be used:
 
 It is highly advised to use at least the `<si-details-pane-header>` to show the back button in a responsive (mobile) view.
 
-When resizing is enabled, the underlying list split part uses the `listUnit`
+When resizing is enabled, the underlying list split part uses the `listWidthUnit`
 input, while the details split part always uses `unit="fr"`. The list pane uses
 `unit="px"` by default. Accordingly, `listWidth` defaults to `300` and
 represents pixels with the default configuration.
 
-Use `listUnit="fr"` to retain a relative split, where `listWidth` is the list
+Use `listWidthUnit="fr"` to retain a relative split, where `listWidth` is the list
 pane's fractional weight. When resizing is disabled, `listWidth` continues to
-control the static layout as a percentage. For both resizable `listUnit="fr"`
+control the static layout as a percentage. For both resizable `listWidthUnit="fr"`
 and static layouts, values outside the inclusive range 0-100 use the standard
 32/68 list/details split.
 
--- a/docs/components/layout-navigation/list-details.md
+++ b/docs/components/layout-navigation/list-details.md
@@ -167,6 +167,10 @@ though it can also be moved to suite the design needs.
 
 <si-docs-component example="si-list-details/si-list-details" height="500"></si-docs-component>
 
+### Default pixel sizing
+
+<si-docs-component example="si-list-details/si-list-details-pixels" height="500"></si-docs-component>
+
 <si-docs-api component="SiListDetailsComponent"></si-docs-api>
 
 <si-docs-api component="SiListPaneComponent"></si-docs-api>
--- a/docs/components/layout-navigation/list-details.md
+++ b/docs/components/layout-navigation/list-details.md
@@ -155,8 +155,9 @@ represents pixels with the default configuration.
 
 Use `listUnit="fr"` to retain a relative split, where `listWidth` is the list
 pane's fractional weight. When resizing is disabled, `listWidth` continues to
-control the static layout as a percentage. Values outside the percentage range
-use the standard 32/68 static layout.
+control the static layout as a percentage. For both resizable `listUnit="fr"`
+and static layouts, values outside the inclusive range 0-100 use the standard
+32/68 list/details split.
 
 If the content may exceed the available space, the `overflow-auto` class should be applied
 to the body components or a child inside them.
--- a/playwright/e2e/element-examples/si-list-details.spec.ts
+++ b/playwright/e2e/element-examples/si-list-details.spec.ts
@@ -104,6 +104,36 @@ test.describe('si-list-details', () => {
     });
   });
 
+  test(`${example} – restore split state after resize`, async ({ page, si }) => {
+    await si.visitExample(example);
+    const split = page.locator('si-split');
+    await expect(split).toHaveCount(1);
+
+    const listPart = page.locator('si-split > si-split-part').first();
+    const initialWidth = (await listPart.boundingBox())!.width;
+
+    // Resize the split by dragging gutter
+    const splitHandle = split.locator('.si-split-gutter');
+    const splitHandleBox = (await splitHandle.boundingBox())!;
+    const x = splitHandleBox.x + splitHandleBox.width / 2;
+    const y = splitHandleBox.y + splitHandleBox.height / 2;
+    await page.mouse.move(x, y);
+    await page.mouse.down();
+    await page.mouse.move(x + 100, y);
+    await page.mouse.up();
+
+    const resizedWidth = (await listPart.boundingBox())!.width;
+    expect(resizedWidth).toBeGreaterThan(initialWidth + 50);
+
+    // Reload the example to verify state restoration
+    await page.goto(page.url() + '?e=' + encodeURIComponent(example));
+
+    // Verify split state (width) is restored
+    const restoredWidth = (await listPart.boundingBox())!.width;
+    expect(restoredWidth).toBeGreaterThan(initialWidth + 50);
+    expect(Math.abs(restoredWidth - resizedWidth)).toBeLessThan(15);
+  });
+
   test('with router in mobile mode', async ({ page, si }) => {
     await page.setViewportSize({ width: 600, height: 800 }); // mdMinimum is 768px
     await si.visitExample('si-list-details/si-list-details-router');
--- a/playwright/e2e/element-examples/si-list-details.spec.ts
+++ b/playwright/e2e/element-examples/si-list-details.spec.ts
@@ -126,12 +126,14 @@ test.describe('si-list-details', () => {
     expect(resizedWidth).toBeGreaterThan(initialWidth + 50);
 
     // Reload the example to verify state restoration
-    await page.goto(page.url() + '?e=' + encodeURIComponent(example));
+    await page.reload();
+    await si.visitExample(example);
+    await expect(listPart).toBeVisible();
 
     // Verify split state (width) is restored
-    const restoredWidth = (await listPart.boundingBox())!.width;
-    expect(restoredWidth).toBeGreaterThan(initialWidth + 50);
-    expect(Math.abs(restoredWidth - resizedWidth)).toBeLessThan(15);
+    await expect
+      .poll(async () => Math.abs((await listPart.boundingBox())!.width - resizedWidth))
+      .toBeLessThan(15);
   });
 
   test('with router in mobile mode', async ({ page, si }) => {
--- a/projects/element-ng/list-details/si-list-details.component.spec.ts
+++ b/projects/element-ng/list-details/si-list-details.component.spec.ts
@@ -14,6 +14,7 @@ import {
   ElementDimensions,
   ResizeObserverService
 } from '../resize-observer';
+import { SiSplitPartComponent, SplitUnit } from '../split';
 import { SiDetailsPaneBodyComponent } from './si-details-pane-body/si-details-pane-body.component';
 import { SiDetailsPaneFooterComponent } from './si-details-pane-footer/si-details-pane-footer.component';
 import { SiDetailsPaneHeaderComponent } from './si-details-pane-header/si-details-pane-header.component';
@@ -37,6 +38,8 @@ import { SiListPaneComponent } from './si-list-pane/si-list-pane.component';
       stateId="si-list-details-1"
       [expandBreakpoint]="expandBreakpoint"
       [disableResizing]="disableResizing()"
+      [listUnit]="listUnit()"
+      [listWidth]="listWidth()"
       [(detailsActive)]="detailsActive"
     >
       <si-list-pane>
@@ -62,6 +65,8 @@ class WrapperComponent {
   readonly listDetails = viewChild.required(SiListDetailsComponent);
   readonly hideBackButton = signal(false);
   readonly disableResizing = signal(true);
+  readonly listUnit = signal<SplitUnit>('px');
+  readonly listWidth = signal(300);
   readonly expandBreakpoint = BOOTSTRAP_BREAKPOINTS.mdMinimum;
   readonly detailsActive = signal(false);
 }
@@ -166,6 +171,8 @@ describe('ListDetailsComponent', () => {
     });
 
     it('should change listWidth when split sizes change', async () => {
+      component.listUnit.set('fr');
+      component.listWidth.set(32);
       component.disableResizing.set(false);
       await fixture.whenStable();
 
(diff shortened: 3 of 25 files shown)


projects/element-ng/list-details/si-list-details.component.ts
```diff
@@ -168,7 +188,23 @@ export class SiListDetailsComponent implements OnInit, OnChanges, OnDestroy {
...
-    this.listWidth.set(sizes[0]);
+    if (this.listUnit() === 'px') {
+      this.listWidth.set(this.getListSizePx(sizes[0]));
+    } else {
+      this.listWidth.set(sizes[0]);
+    }
+  }
+
+  private getListSizePx(fallbackPercentage: number): number {
+    const listPart = this.elementRef.nativeElement.querySelector<HTMLElement>('si-split-part');
```
chintankavathia: use viewChild instead

projects/element-ng/list-details/si-list-details.component.ts
```diff
@@ -168,7 +188,23 @@ export class SiListDetailsComponent implements OnInit, OnChanges, OnDestroy {
...
+    const listWidth = listPart?.getBoundingClientRect().width;
+    if (listWidth) {
+      return listWidth;
+    }
+
+    const split = this.elementRef.nativeElement.querySelector<HTMLElement>('si-split');
```
chintankavathia: same here. use viewChild

projects/element-ng/list-details/si-list-details.component.ts
```diff
@@ -105,10 +115,20 @@ export class SiListDetailsComponent implements OnInit, OnChanges, OnDestroy {
...
   readonly stateId = input<string>();
 
-  protected readonly splitSizes = computed<[number, number]>(() => [
-    this.listWidth(),
-    100 - this.listWidth()
-  ]);
+  /** @internal */
+  readonly staticListWidth = computed(() => {
+    const listWidth = this.listWidth();
+    return listWidth >= 0 && listWidth <= 100 ? listWidth : 32;
```
chintankavathia: 32 should be declared as named constant or atleast have a comment about it

chintankavathia (changes requested):

chintankavathia (approved):
@kfenner Please take a look at the commit message as I am not sure if we go with ! format.

projects/element-ng/list-details/si-list-details.component.ts
```diff
@@ -168,7 +196,21 @@ export class SiListDetailsComponent implements OnInit, OnChanges, OnDestroy {
...
+
+  private getListSizePx(fallbackPercentage: number): number {
+    const listWidth = this.listPart()?.nativeElement.getBoundingClientRect().width;
+    if (listWidth) {
+      return listWidth;
+    }
+
+    const splitWidth = this.split()?.nativeElement.getBoundingClientRect().width;
+    return splitWidth ? (splitWidth * fallbackPercentage) / 100 : fallbackPercentage;
   }
```
kfenner, line 221: This is the part I'd most like to see changed. `getMainSizePx` in #2728 solves the exact same problem one commit earlier:

```ts
private getMainSizePx(fallbackPercentage: number): number {
  const mainPartSize = this.mainSplitPart()?.expandedSize();
  if (mainPartSize !== undefined) {
    return Math.round(mainPartSize);
  }
  const splitWidth = this.splitElement()?.nativeElement.getBoundingClientRect().width;
  return splitWidth ? Math.round((splitWidth * fallbackPercentage) / 100) : this.minMainSize();
}
```

Four differences, and I think #2728 is better on each:

1. **Forced synchronous reflow on every `mousemove`.** `sizesChange` fires for each drag step, so this calls `getBoundingClientRect()` (and a second one in the fallback) on every pointer move. `SiSplitPartComponent` already tracks the pixel size in `expandedSize()` — the split component computed it, converted it to a percentage to emit, and here we go back to the DOM to recover the pixels we already had. Reading `expandedSize()` avoids the layout thrash entirely and removes the lossy px → % → measure → px round trip.

2. **No rounding.** `listWidth` is a `model`, so consumers get values like `399.328125` on every drag step. #2728 rounds.

3. **`if (listWidth)` treats a legitimate `0` as "not measured".** `expandedSize() !== undefined` is the precise check.

4. **Unit bug in the final fallback.** `return splitWidth ? ... : fallbackPercentage;` — when the split element can't be measured, this returns a 0–100 percentage into a field that is documented as pixels, so `listWidth` would be set to e.g. `40` (px). #2728 falls back to `minMainSize()`, which at least is a plausible pixel value. `minListSize` would be the equivalent here.

Was there a reason to measure the DOM rather than read `expandedSize()`, e.g. staleness in some code path I'm not seeing? If not, I'd just mirror #2728.

projects/element-ng/list-details/si-list-details.component.ts
```diff
@@ -41,9 +43,15 @@ import { BehaviorSubject, Subject, Subscription } from 'rxjs';
...
-  private elementRef = inject(ElementRef);
+  private elementRef = inject<ElementRef<HTMLElement>>(ElementRef);
   private resizeObserver = inject(ResizeObserverService);
   private readonly listDetailsContainer = viewChild.required<ElementRef>('listDetailsContainer');
+  private readonly listPart = viewChild(SiSplitPartComponent, {
+    read: ElementRef<HTMLElement>
+  });
+  private readonly split = viewChild(SiSplitComponent, {
+    read: ElementRef<HTMLElement>
+  });
```
kfenner: `viewChild(SiSplitPartComponent, ...)` resolves to the *first* split part positionally, so this silently binds to the wrong element if the template order ever changes. #2728 used an explicit template reference for the same lookup (`viewChild('mainSplitPart', { read: SiSplitPartComponent })`), which I'd prefer here too — especially since (per the other comment) reading `SiSplitPartComponent` instead of `ElementRef` would also let you use `expandedSize()`.

projects/element-ng/list-details/si-list-details.component.ts
```diff
@@ -105,10 +123,20 @@ export class SiListDetailsComponent implements OnInit, OnChanges, OnDestroy {
...
+
+  protected readonly splitSizes = computed<[number, number]>(() => {
+    if (this.listUnit() === 'fr') {
+      const relativeListWidth = this.staticListWidth();
+      return [relativeListWidth, 100 - relativeListWidth];
+    }
+
+    return [this.listWidth(), 1];
+  });
```
kfenner: Two things here:

**Naming.** `staticListWidth` isn't only used for the static layout — the `fr` branch of `splitSizes` below uses it as the resizable size too, and `fr` is by definition the non-static case. #2728 keeps these apart in `getSplitSizes()` with distinct fallbacks per branch (`[50, 50]` static, `[32, 68]` resizable-`fr`). Either rename it to something unit-neutral (`listWidthPercent`?) or split the two concerns like #2728 does.

**Silent clamping.** With `listUnit="fr"`, any `listWidth` outside 0–100 silently snaps to 32 with no dev-mode warning. This is load-bearing for the migration — a component with no explicit `listWidth` gets the new default `300`, which is out of range, falls back to 32 and thereby restores the old 32/68 split. That's clever, but it's entirely implicit; if the range check is ever loosened the migration quietly changes behaviour. Worth at least a comment explaining the dependency.

projects/element-ng/list-details/si-list-details.component.ts
```diff
@@ -79,11 +87,21 @@ export class SiListDetailsComponent implements OnInit, OnChanges, OnDestroy {
...
   /**
-   * The percentage width of the list view of the overall component width.
+   * The initial size of the list split part when resizing is enabled.
+   * The value uses {@link listUnit}. With `listUnit="fr"`, the value is
+   * treated as a percentage-like fractional weight. In the static layout,
+   * numeric values continue to represent a percentage.
+   *
+   * @defaultValue 300
+   */
+  readonly listWidth = model<number>(300);
```
kfenner, line 99: The JSDoc is thinner than the equivalent on `mainContainerWidth` in #2728, which spells out both the 0–100 constraint and the out-of-range fallback. Here "percentage-like fractional weight" doesn't tell me that values outside 0–100 are silently replaced by 32.

Also, "The initial size" understates it: this is a `model` that is written back on every resize, so it's the current size, not just the initial one.

projects/element-ng/schematics/ng-update/migrate-list-details-units.ts
```diff
@@ -0,0 +1,152 @@
...
+
+  const hasListUnit = element.attrs.some(attr => listUnitAttributeNames.includes(attr.name));
+  if (hasListUnit) {
+    return;
+  }
+
+  insertAttribute(template, element, 'listUnit="fr"', offset, recorder);
+};
+
+const insertAttribute = (
```
kfenner: `insertAttribute` is already exported from `projects/element-ng/schematics/utils/template-edit-utils.ts` (re-exported via `../utils/index.js`), and `migrate-main-detail-units.ts` — added in #2728, in this same stack — imports it from there. This local copy looks byte-identical, so it can just be deleted and added to the existing import block at the top of the file.

Separate nit, not for this PR: `getStaticBooleanValue` and `getExpression` now exist in four near-identical copies across `migrate-list-details-units.ts`, `migrate-main-detail-units.ts`, `migrate-split-collapse.ts` and `migrate-split-scale.ts`. Might be worth hoisting into `utils` as a follow-up.

playwright/e2e/element-examples/si-list-details.spec.ts
```diff
@@ -104,6 +104,36 @@ test.describe('si-list-details', () => {
...
+    const resizedWidth = (await listPart.boundingBox())!.width;
+    expect(resizedWidth).toBeGreaterThan(initialWidth + 50);
+
+    // Reload the example to verify state restoration
+    await page.goto(page.url() + '?e=' + encodeURIComponent(example));
+
+    // Verify split state (width) is restored
+    const restoredWidth = (await listPart.boundingBox())!.width;
+    expect(restoredWidth).toBeGreaterThan(initialWidth + 50);
+    expect(Math.abs(restoredWidth - resizedWidth)).toBeLessThan(15);
```
kfenner: I don't think this test can fail, for two independent reasons:

1. **It doesn't reload.** `visitExample` navigates to `#/si-list-details/si-list-details` (the `?e=` form is only used for the live-previewer projects). So `page.url() + '?e=' + ...` yields `.../#/si-list-details/si-list-details?e=si-list-details%2Fsi-list-details` — a URL that differs from the current one *only in the fragment*. That's a same-document navigation, so no document reload happens and the assertions below just re-measure the DOM that was already resized. The double `?e=` would also be a malformed route if it did navigate.

2. **Nothing is persisted anyway.** `SiSplitComponent` does `inject(SI_UI_STATE_SERVICE, { optional: true })` and no-ops in both `saveUIState` and `restoreFromUIState` when it's absent. The demo app never calls `provideSiUiState()`, so `stateId="si-list-details-demo"` has no effect there.

So as written this asserts "the width I just dragged to is still the width". To make it meaningful you'd need `await page.reload()` (or a fresh `visitExample`) *and* a UI state provider in the examples app. If wiring up persistence is out of scope, I'd drop the restore part and keep the resize assertions, rather than leave a test that looks like it covers persistence.

src/app/examples/si-list-details/si-list-details.html
```diff
@@ -2,6 +2,7 @@
   #container
   stateId="si-list-details-demo"
   class="p-6 px-md-9"
+  listUnit="fr"
```
kfenner: Both examples were migrated to `listUnit="fr"`, which is correct for preserving their appearance — but it means the new default (`listUnit="px"`, `listWidth` 300) has no example and therefore no e2e/VRT coverage at all, even though it's the behaviour every consumer gets after upgrading. Could one of the two examples switch to the new default instead, or a third small one be added? The `px` path is also the only one that exercises `getListSizePx`.

projects/element-ng/list-details/si-list-details.component.spec.ts
```diff
@@ -183,6 +190,68 @@ describe('ListDetailsComponent', () => {
...
+      expect(parts[1]!.componentInstance.unit()).toBe('fr');
+      expect(parts[1]!.componentInstance.size()).toBe(60);
+    });
+
+    it('should keep percentage sizing for the static layout', async () => {
+      component.listWidth.set(300);
+      await fixture.whenStable();
+
+      expect(getListPane().style.flexBasis).toBe('32%');
+      expect(getDetailsPane().style.flexBasis).toBe('68%');
```
kfenner: The test name says "keep percentage sizing", but `300` is out of range so this only asserts the 32/68 *fallback*. The actual regression risk — that an in-range percentage still passes through to the static layout — isn't covered. Worth adding a case like `listWidth.set(40)` → `40%` / `60%`.

docs/components/layout-navigation/list-details.md
```diff
@@ -148,6 +148,16 @@ Inside the `<si-details-pane>` the following child components can be used:
...
 It is highly advised to use at least the `<si-details-pane-header>` to show the back button in a responsive (mobile) view.
 
+When resizing is enabled, the underlying list split part uses the `listUnit`
+input, while the details split part always uses `unit="fr"`. The list pane uses
+`unit="px"` by default. Accordingly, `listWidth` defaults to `300` and
+represents pixels with the default configuration.
+
+Use `listUnit="fr"` to retain a relative split, where `listWidth` is the list
+pane's fractional weight. When resizing is disabled, `listWidth` continues to
+control the static layout as a percentage. Values outside the percentage range
```
kfenner: "When resizing is disabled, `listWidth` continues to control the static layout as a percentage. Values outside the percentage range use the standard 32/68 static layout."

The 32/68 fallback also applies with resizing *enabled* when `listUnit="fr"` (see `staticListWidth` feeding the `fr` branch of `splitSizes`), so this reads as more narrowly scoped than it is.

Related: the PR description says "Static layouts (`disableResizing="true"`) are unaffected", but they aren't entirely — previously an out-of-range `listWidth` was passed straight through to `flex-basis` (e.g. `150%`), and now it clamps to `32%`. Probably an improvement, but since that text becomes the changelog entry it'd be good to have it match.

kfenner (commented):
Reviewed as the last PR of the stack, mostly by diffing it against #2728, which solves the same problem for `si-main-detail-container`. The overall approach and the migration semantics look right to me — in particular, migrating `si-list-details` unconditionally (unless `disableResizing` is statically `true`) is correct here because resizing is opt-out, whereas #2728 is opt-in via `resizableParts`. Nice catch.

My main concern is that the pixel-readback helper and the split-size computation were re-invented here instead of following the shape that #2728 established one commit earlier, and the new implementation is measurably weaker. Details inline.

Summary:

- **`getListSizePx` should follow `getMainSizePx` from #2728** — it forces a synchronous layout on every `mousemove`, doesn't round, and its last fallback returns a percentage as if it were pixels. (main comment)
- **`insertAttribute` is duplicated** in the new schematic instead of imported from `../utils/index.js`, which the sibling migration in this same stack already does.
- **The new e2e test can't actually fail** — the "reload" is a fragment-only `page.goto` (no document reload), and the examples app never provides `SI_UI_STATE_SERVICE`, so nothing is persisted or restored.
- **No example or e2e covers the new default `px` path** — both examples were pinned to `listUnit="fr"`.
- Some naming/doc gaps around `staticListWidth` and the silent 32/68 fallback.

None of these are blocking as far as I'm concerned, but the first three feel worth fixing before merge.

timowolf (approved):

projects/element-ng/list-details/si-list-details.component.ts
chintankavathia: fixed

projects/element-ng/list-details/si-list-details.component.ts
chintankavathia: fixed

projects/element-ng/list-details/si-list-details.component.ts
chintankavathia: doc updated

projects/element-ng/schematics/ng-update/migrate-list-details-units.ts
chintankavathia: removed

projects/element-ng/list-details/si-list-details.component.spec.ts
chintankavathia: updated tests

docs/components/layout-navigation/list-details.md
chintankavathia: updated docs

playwright/e2e/element-examples/si-list-details.spec.ts
chintankavathia: updated e2e.

src/app/examples/si-list-details/si-list-details.html
chintankavathia: new example added for pixel units

projects/element-ng/list-details/si-list-details.component.ts
chintankavathia: added expandedSize based read

kfenner (approved):
LGTM 👍

kfenner (approved):

kfenner (approved):
LGTM 👍

PR #2742, opened 2026-09-08
refactor(element-theme): deprecate list group

kfenner:
The Bootstrap based list group is superseded by the [Element list](https://element.siemens.io/development/components/lists-tables-trees/list/). This removes it from the documentation and marks the code as deprecated, without breaking anything.

## What changed

**Docs**

- Removed `docs/components/lists-tables-trees/list-group.md`, its image and the nav/index entries.
- Retargeted the links in `lists-tables-trees/overview.md` and `status-notifications/circle-status.md` to the list page.
- Added a *Migrating from the list group* section with a full mapping table to `list.md`.

**Code**

- `bootstrap/_list-group.scss`: file level `@deprecated` header with the migration mapping, plus a `@deprecated` note on every top level selector.
- `bootstrap/_variables.scss`: `@deprecated` note on the `$list-group-*` block.

No selectors or values were changed, so this is purely additive from a runtime perspective. The `list-group/list-group` example, its Playwright static test and the VRT snapshots stay in place, the styles are still supported and stay covered, they are just no longer recommended.

## Why no schematics migration

`applyElementClassMigration` can only rewrite `class` attributes. That is not enough here:

- `.list-item` is a CSS grid with named areas and a `> * { grid-column: 3 / -1 }` catch all, so renaming `.list-group-item` to `.list-item` would put every existing child on its own row and silently break the layout.
- Most legacy classes have no equivalent at all (`.list-group-md`/`-lg`, `.list-group-horizontal*`, `.list-group-numbered`, `.list-group-flush`, the `.list-group-item-*` color variants, `.list-header`).

Migration needs per call site judgement, so it is documented instead of automated. The mapping table lives both in the docs and in the `DEPRECATED:` commit footer so it lands in the changelog.

## Removal

Deliberately open ended. The styles stay available, there is no removal planned at this point, and the wording keeps the option to remove them in a future major version.

## Follow ups, not in this PR

- `docs/components/layout-navigation/cards.md` still shows `list-group list-group-flush` samples. Cards have dedicated list group SCSS in `_card.scss`, so moving those to `.list` needs card styling work first.
- Internal usages keep working and were left alone: `si-about`, `si-empty-state`, `si-list-widget`, `si-widget-catalog` and a few examples.

---

[Documentation](https://d33c9dcnqinn2a.cloudfront.net/pr-2742/pages/).
[Examples](https://d33c9dcnqinn2a.cloudfront.net/pr-2742/pages/element-examples/).
[Dashboards Demo](https://d33c9dcnqinn2a.cloudfront.net/pr-2742/pages/dashboards-demo/).
[Playwright report](https://d33c9dcnqinn2a.cloudfront.net/pr-2742/playwright-report/).

**Coverage Reports:**

Code Coverage

- [charts-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-2742/pages/coverage/charts-ng/)
- [dashboards-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-2742/pages/coverage/dashboards-ng/)
- [element-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-2742/pages/coverage/element-ng/)
- [element-translate-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-2742/pages/coverage/element-translate-ng/)
- [maps-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-2742/pages/coverage/maps-ng/)
- [native-charts-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-2742/pages/coverage/native-charts-ng/)

## Known gaps blocking full migration

Two capability gaps in the replacement component mean a handful of internal consumers cannot migrate yet, tracked in #2748:

- `.list-item` has no persistent selected/current state (only the transient `:active` pressed state), which blocks every listbox / selection pattern.
- `.list-item-action` on an `<a>` sets `color: inherit`, so link colour depends on the consumer also adding `text-primary`.

This does not affect the deprecation itself — the list group remains fully functional — but it is why `si-widget-catalog`, `si-list-widget-*` and `custom-widget-catalog` still use it.


dr-itz (approved):

projects/element-theme/src/styles/bootstrap/_list-group.scss
chintankavathia, line 1: shouldn't we rename the whole file with deprecated suffix?. `_list-group-deprecated.scss`

kfenner: @chintankavathia Did we do this in other cases?

kfenner: It would mean people have to change paths...

kfenner: I want this to be low-impact, ideally.

dr-itz: can do, this is one additional line of change in the import. Consumers have no business importing this directly. Sure they could, but that's kinda pointless when it is included in the theme anyway.

chintankavathia: Its very unlikely that someone would be explicitly importing this path and even in that case we could just add schematic to handle it.

chintankavathia:
Looking at this together with #2745 and #2746, card integration and the compatible internal migrations are covered. Since the legacy CSS stays available, the points below are migration/parity concerns, not immediate regressions for existing consumers.

### Remaining capability gaps

- **Persistent selected/current row:** The new list has the transient `:active` pressed effect but no built-in equivalent to `.list-group-item.active`. This matters for master-detail selection and widget catalogs. This is already acknowledged and tracked in #2748; it is not a new finding against the deprecation itself.
- **Disabled rows:** The old `.disabled`/`:disabled` rules provide disabled styling and suppress pointer interaction. The replacement has no equivalent. Native `<button disabled>` still prevents activation, but retains the normal list hover/cursor styling. Anchors need explicit disabled styling and interaction handling; `aria-disabled` alone does not prevent activation.
- **Numbered rich lists:** The recommendation to use an ordered list does not preserve numbering when retaining `.list-item`: `ol:has(.list-item)` sets `list-style: none`, and `.list-item` uses grid layout. This needs a working counter/marker example or an explicit statement that there is no equivalent.

#2748 also tracks the anchor-color default: navigation still works, but `.list-item-action` uses `color: inherit`, so consumers currently need an explicit color utility such as `text-primary` for the intended link appearance.

### Migration-guide corrections

1. [Action mapping](https://github.com/siemens/element/blob/5d1920c15ecc9bf104ff7e3ba4a8af868b034428/docs/components/lists-tables-trees/list.md#L121): put **both** `.list-item` and `.list-item-action` on the same `<button>` or `<a>`, wrapped in an `<li>`. Putting only `.list-item-action` inside a separate `.list-item` does not match the action styles. The existing action example already uses the correct structure.
2. [Flush mapping](https://github.com/siemens/element/blob/5d1920c15ecc9bf104ff7e3ba4a8af868b034428/docs/components/lists-tables-trees/list.md#L122): `.list.list-divider` is the closer replacement when preserving separators. Plain `.list` removes the internal separators as well as the outer border. The follow-up card examples already use `.list.list-divider`.
3. [Numbered mapping](https://github.com/siemens/element/blob/5d1920c15ecc9bf104ff7e3ba4a8af868b034428/docs/components/lists-tables-trees/list.md#L125): clarify the limitation above rather than suggesting `<ol>` alone is sufficient.

The legacy minimum-height variants (`.list-group-md`/`-lg`), responsive horizontal groups, contextual row colors, and styled group headers also have no built-in equivalents. These are missing convenience styles rather than unsupported content, but concrete utility/custom-CSS migration examples would help.

Suggested scope here: correct the action/flush/numbering guidance and explicitly document the selected/disabled limitations, while keeping the missing state support as follow-up work. The staged deprecation can remain non-breaking without claiming complete replacement parity.

projects/element-theme/src/styles/bootstrap/_list-group.scss
kfenner: let's leave the file as is, I think we didn't do this for .css stuff so far.

kfenner:
Addressed the three migration-guide corrections in a029934f4:

- `.list-item-action` is now documented on the same `<button>` or `<a>` as `.list-item`.
- `.list.list-divider` is documented as the flush replacement when dividers should remain.
- Numbered lists now explicitly require a custom counter because `.list-item` removes list markers.

The selected and disabled points are intentional scope differences from the Bootstrap list group, not blockers for this non-breaking deprecation. Element lists use explicit controls for selection; interaction-specific patterns such as `cdkListbox` own their selected and disabled state and styling. Existing list-group styles remain available for consumers that need those legacy affordances.

kfenner:
@chintankavathia @dr-itz I'm happy with this now. I think it's good to merge. If you are good too, feel free to press the button 🙏 

We should NOT merge #2745 and #2746 yet, we first need to clean them up and make sure all changes are as intended. Also we might want to add back later some support for `-sm` and `-lg` list items etc.

chintankavathia:
> @chintankavathia @dr-itz I'm happy with this now. I think it's good to merge. If you are good too, feel free to press the button 🙏
> 
> We should NOT merge #2745 and #2746 yet, we first need to clean them up and make sure all changes are as intended. Also we might want to add back later some support for `-sm` and `-lg` list items etc.

Overall looks good but since we have migration chapter in list docs we should also add small snippet of before and after for very basic example additionally current commit message code example doesn't migrate the given example correctly. 

currently it is:

Before:
```
    <ul class="list-group">
      <li class="list-group-item">Item</li>
    </ul>
```
After:
```
    <ul class="list">
      <li class="list-item">
        <span class="list-item-title">Item</span>
      </li>
    </ul>
```

but it should be:
```
<ul class="list list-divider background-1 rounded-2"> 
    <li class="list-item">Item</li>
</ul>
```

kfenner:
@panch1739 Thanks for the very helpful migration guidance image 🙇 

@chintankavathia This should be ready to merge 🙏

chintankavathia (approved):
👍

dr-itz (approved):

timowolf (approved):

PR #2745, opened 2026-09-08
feat(element-theme): support element list inside cards

kfenner:
Follow-up to #2742 (base branch `kaspar-fenner-deprecate-list-group`). That PR deprecates the Bootstrap `.list-group`; this one makes the Element `.list` a working replacement inside `.card`, and migrates the remaining examples and docs.

The library change is **purely additive** — no `.list-group` rule is removed or altered, so existing markup is unaffected.

## `_card.scss`

Two things were missing for `.list` in a card.

**1. Rounded corners.** A `.list` as first/last child now gets the card's inner radius, and the outer corners of the first/last item inherit it, so the background of a hovered, filled or outlined item no longer paints over the card's rounded corners.

The per-corner rules are written as `> :first-child { &, > .list-item { ... } }` because `.list-item` sits either on the `<li>` itself or on a wrapped `<button class="list-item list-item-action">`. Specificity works out: `.card > .list > :first-child > .list-item` is (0,4,0) and beats `.list.list-filled .list-item` at (0,3,0).

**2. Focus ring.** `.list` was added to the existing `.card-header + :is(...)` rule that switches the focus outline to an inset offset. That half alone is a regression, so it is paired with the `z-index` treatment that `.list-group` already has — extending the precedent from ec3d92f9fc (`fix(theme): fix card spacing for focus outline`).

### Focus-visibility bug

A scrollable list following a `.card-header` draws its focus ring *inside* itself. Without the pairing, any item background paints over it. This is not a filled/outline edge case — plain `.list-divider` items are transparent at rest but paint `$si-sys-background-hover` on hover, and hovering the top row of a scrollable list in a card is ordinary interaction.

`.list-divider`, first item hovered — before / after:

 

`.list-filled`, focused — before / after (previously the ring survived only in the `gap` between items):

 

Verified across all six states (`list-divider`/`list-filled`/`list-outline`, each focused and focused+hovered) — all show an unbroken ring after the change.

Note the rule targets `> li` rather than `.list-item`. `z-index` only applies to positioned elements and flex items; in a `.list` the flex items are the `<li>` children, so a `.list-item` rule would silently no-op for the wrapped-button shape. Pushing the `li` back takes any wrapped button with it. `z-index: 0` on the container scopes the negative children to the list instead of letting them sink behind the card background.

## `si-card` is not affected

The Angular `si-card` component projects `[body]` into a `.content-container` that already sets `overflow: hidden; border-radius: inherit`, so `.card > .list` never matches projected content. Its header is `<si-card-header>`, not `.card-header`. These rules only affect native `.card` markup.

## Examples

Five examples still used `.list-group` and were migrated. Slot mapping: `.list-item-indicator` replaces the `my-n4 me-5` / `ps-4` spacing hacks, `.list-item-title` carries h5 typography (so `si-h5` is dropped), `.list-item-primary-action` replaces manual action alignment.

Two needed more than a class swap — both had long runs of hand-duplicated markup, partly mangled by prettier into the `</li\n>` form, which is not reasonably editable by hand:

- `si-fixed-height-layout-side-panel` — 16 near-identical `<li>` blocks → `@for` over a `listItems` array
- `si-list-widget-css` — 6 blocks → `@for` over a `buildings` array

These are template restructures, not scope creep. In `si-list-widget-css` the rows were all literally labelled "Building B"; they are now B–G. That example is not VRT-covered, so the rename is safe.

`si-weather-widget-configurable` also had invalid HTML — a `<button>` as a direct child of `<ul>` — now wrapped in an `<li>`.

## Docs

`cards.md` — all three card+list snippets and surrounding prose rewritten; no `list-group` references remain. `list.md` — new "List in a card" section. The migration table from the base branch is deliberately untouched.

## Verification

- Dev server compiles the demo app cleanly (covers the `@for` conversions and both `.ts` changes)
- `stylelint` and `prettier` clean over all touched files
- Focus/radius behaviour verified by screenshot across list variants

VRT snapshots are **not** regenerated in this PR — only `si-card/si-card` and `si-fixed-height-layout-side-panel` are snapshot-covered, and both change visually as intended. They need a run in the Docker image.

---

[Documentation](https://d33c9dcnqinn2a.cloudfront.net/pr-2745/pages/).
[Examples](https://d33c9dcnqinn2a.cloudfront.net/pr-2745/pages/element-examples/).
[Dashboards Demo](https://d33c9dcnqinn2a.cloudfront.net/pr-2745/pages/dashboards-demo/).
[Playwright report](https://d33c9dcnqinn2a.cloudfront.net/pr-2745/playwright-report/).

**Coverage Reports:**

Code Coverage

- [charts-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-2745/pages/coverage/charts-ng/)
- [dashboards-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-2745/pages/coverage/dashboards-ng/)
- [element-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-2745/pages/coverage/element-ng/)
- [element-translate-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-2745/pages/coverage/element-translate-ng/)
- [maps-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-2745/pages/coverage/maps-ng/)
- [native-charts-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-2745/pages/coverage/native-charts-ng/)


kfenner:
@panch1739 Could you check whether these visual changes are actually what we want?

Two existing examples were migrated from `.list-group` to `.list`, and the screenshots changed. I would have expected them to stay the same.

**si-card** — the scrollable list next to the image. Item padding and wrapping changed, and the list now follows the card corner.

Before:

si-card before

After:

si-card after

**Side panel** — the list used to be inset (`ms-6 rounded-0 border-0`). It is now flush with the card. `ms-6` was removed because `.list-item` padding already matches the card header, but that is a layout change, not just a class rename.

Before:

side panel before

After:

side panel after

If these should stay as they were, the migration notes need to say how to keep the old look (extra margin, `rounded-0`) instead of treating the new card-list styling as the default replacement.

Which of these is correct?

1. The new look is the intended replacement.
2. Existing examples should stay visually the same, and the migration guide should spell out how.

kfenner:
@panch1739 follow-up — the earlier comment missed the main visual change.

Besides the padding, the item text had become **bold**. That was not intended. The migration examples wrapped the old plain text in `.list-item-title`, and the guide made that slot look required. `.list-item-title` is only the heading slot: it is semibold and adds its own block padding. A plain migration does not need it.

The guide and examples are corrected so the old text stays regular weight. Please ignore the previous "after" images; they still show that accidental bold.

What is still different, and what I would like you to judge:

1. List-item padding versus the old list-group padding (the card list wraps a little differently).
2. A list that meets a card edge now inherits that edge's radius.
3. The side panel no longer uses `ms-6 rounded-0 border-0`, so the list is flush with the card instead of inset.

Updated after, light:

si-card after, regular weight

side panel after, regular weight

PR #2763, opened 2026-09-12
docs: add update guide for v51

kfenner:
Adds a v51 update guide. The v49 guide stays in the nav for users coming from v48 or older.

The guide covers:

- Angular 22 and peer dependency bumps
- What the `migration-v51` schematic applies (collapsed; expand to see the list)
- Remaining API changes that need a compile-error or visual check
- Migrating deprecated `$element-*` tokens to `$si-sys-*`, with mapping JSON files and an example AI prompt

It stays terse and links to the changelog rather than restating it. The release notes and the upcoming blog post cover the "what's new" angle.

There is no Element v50.

### Notes for reviewers

One detail was verified against the schematic source rather than the changelog:

- The `text-primary` to `text-accent` and `text-body` to `text-primary` renames are applied against a snapshot of the original classes, so they cannot collide within a schematic run. A manual find-replace after the update would apply them a second time and turn body text into accent text. The guide therefore scopes the manual renames to the places the schematic does not rewrite -- stylesheets, class bindings and TypeScript strings -- and spells out the order for the two swapped pairs.

The color mapping file had one typo fixed against the token source (`si.sys.border.border.1` → `si.sys.border.1`). Descriptions no longer mention internal package names.

`mkdocs build --strict` passes with no warnings on the new page.

---

[Documentation](https://d33c9dcnqinn2a.cloudfront.net/pr-2763/pages/).
[Examples](https://d33c9dcnqinn2a.cloudfront.net/pr-2763/pages/element-examples/).
[Dashboards Demo](https://d33c9dcnqinn2a.cloudfront.net/pr-2763/pages/dashboards-demo/).
[Playwright report](https://d33c9dcnqinn2a.cloudfront.net/pr-2763/playwright-report/).

**Coverage Reports:**

Code Coverage

- [charts-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-2763/pages/coverage/charts-ng/)
- [dashboards-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-2763/pages/coverage/dashboards-ng/)
- [element-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-2763/pages/coverage/element-ng/)
- [element-translate-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-2763/pages/coverage/element-translate-ng/)
- [maps-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-2763/pages/coverage/maps-ng/)
- [native-charts-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-2763/pages/coverage/native-charts-ng/)


docs/update-guides/update-v51.md
```diff
@@ -0,0 +1,126 @@
+# Update to Element v51
+
+Element v51 requires Angular 22 and changes the visual appearance of every component.
+The schematics handle most template and token rewrites. API removals still fail the build.
```
kfenner: I don't think it is correct to say that the visual appearance of every component changes. Mostly the visual appearance should stay more or less the same. We are using new design system tokens from the Siemens Design Language and replace the old $element tokens (they are still available, just deprecated). From where is this sentence/conclusion coming from?

And what does "API removals still fail the build" mean? Schematics handle most breaking changes, no? It's a strange introduction for this release tbh.

docs/update-guides/update-v51.md
```diff
@@ -0,0 +1,126 @@
...
+- `si-icon-status` replaced by `si-status-counter`
+- Removed `provideIconConfig()` and the removed inputs `showLessAppsText`, `uploadTextFileSelect` and `tabbable`
+- `si-map` `moreText` moved to `si-map-tooltip`
+- `--si-feedback-icon-offset` replaced by `--si-feedback-icon-size`
+
+Still using the legacy tabs? Migrate them with:
+
+```sh
+ng g @siemens/element-ng:migrate-tabs-legacy
+```
```
kfenner: double check if this list is conclusive or shortened to the most important points, would be good to know.
The language used here is short, which is good, but it's not 100% consistent across all bullet points, the intro sentencea bove says ".. schematics cover:" but e.g. the first sentence doesn't say what is actually being done with the typography variables, are they migrated from old to new or  replaced, removed, whatever, it's not clear. Check the exact wording for all the bullet points and make sure they are understandable and semantically correct.

docs/update-guides/update-v51.md
```diff
@@ -0,0 +1,126 @@
...
+
+### Utility classes the schematic cannot see
+
+The schematic already rewrote static `class` attributes. A template-wide or workspace-wide
+replace inverts that work (`text-primary` and `btn-ghost` change meaning a second time).
+
+Two class pairs were repurposed. In stylesheets and in `[ngClass]`, `[class]` and
+`[class.*]` values only, apply these renames once, in this order:
```
kfenner: I don't understand this whole section. Did the schematic do it's job or not? Do I have to do some stuff manually? What is this whole thing about, super confusing.

docs/update-guides/update-v51.md
```diff
@@ -0,0 +1,126 @@
...
+| Sort bar | `SiSortBarComponent` removed without replacement. |
+| Datepicker | `calenderWeekLabel` renamed to `calendarWeekLabel`. |
+| Filtered search | `datepickerConfig` no longer accepts `enableDateRange` and `enableTwoMonthDateRange`. Remove them. |
+| Status bar | `expandButtonText` and `collapseButtonText` replaced by `toggleButtonText`. Translation key: `SI_STATUS_BAR.TOGGLE`. |
+| Loading spinner | `LOADING_SPINNER_BLOCKING` and `LOADING_SPINNER_OVERLAY` removed. Use the `isBlockingSpinner` and `isSpinnerOverlay` inputs. |
+| Gauge chart | Use `valueFormatter` to format the value. `labelFormatter` now only formats axis labels. |
+| Dashboards | `SiWidgetCatalogComponent.closed` now emits an array. |
+| Maps | `MapPoint.extraProps` replaced by `extraProperties`. `DEFAULT_FIT_PADDING` removed. |
+| Resize observer | `ResizeObserverService._checkAll` removed. Use `mockResizeObserver()` in tests. |
+| Theme | `$spacers` values are now CSS variables and need `calc()` in math expressions. |
```
kfenner: are these things not handled by the schmeatics or what? "Apply the remaining changes" reads a bit like that, so, all manual work now??? please double check. If true, is there a reason why schematics cannot handle it?

docs/update-guides/update-v51.md
```diff
@@ -0,0 +1,126 @@
...
+
+### Formly
+
+`@siemens/element-ng/formly` now requires ngx-formly v7.1. If you stay on v6, switch your
+imports to `@siemens/element-ng/formly-legacy` and keep your current ngx-formly version.
+See [dynamic forms](../architecture/dynamic-forms.md).
```
kfenner: why is this a separate section? If it's needs to be, ok, but just found it strange.

docs/update-guides/update-v51.md
```diff
@@ -0,0 +1,126 @@
...
+
+## Review the UI
+
+Run your application and check every screen. Nothing below fails the build.
```
kfenner: strange sentence somehow!? I guess you mean there are a few things which need extra care etc. maybe explain this a bit friendlier?

docs/update-guides/update-v51.md
kfenner: Fair, it was confusing. The section is about a real gap: the class-name rules only rewrite the static `class=...` attribute in templates (`element-class-change.migration.ts`). Class names anywhere else are left untouched, so yes, there is manual work, but only in those places.

The old text also blurred class names and stylesheet variables. Custom properties and Sass variables *are* rewritten inside your stylesheets. It now says that in the first paragraph:

> The schematic rewrites class names only in the static `class` attribute of your templates â€” not in stylesheets, in `[ngClass]`, `[class]` and `[class.*]` bindings, or in strings you build in TypeScript. CSS custom properties and Sass variables in your stylesheets are already rewritten.

Then it says what to do: rename the leftovers with the same mappings, and do the two swapped pairs in the given order. The warning against a project-wide find and replace stays, because the schematic snapshots the original classes once per element. Re-running the swaps over already-migrated templates would turn your body text into accent text.

docs/update-guides/update-v51.md
kfenner: Double-checked all 13 rows against the schematic sources: none of them are covered. The migration data for attribute selectors, class member replacements and component property names is empty, so nothing in that table is touched by `ng update`.

On saying why: I tried it and every phrasing I came up with was either hand-wavy or falsifiable (the `calenderWeekLabel` typo, for instance, could have been automated). The honest answer differs per row: no mechanical 1:1 equivalent, removed without replacement, or it needs a decision only you can make. That is three sentences of meta-explanation the reader cannot act on, so I left it out.

What the reader does need is whether the build will catch it. The intro now says that:

> The schematic does not cover the changes below. They show up as build errors, except File uploader and Gauge chart:

docs/update-guides/update-v51.md
kfenner: Agreed, it did not earn a section. Folded into the peer dependency table as a row plus one short paragraph right below it:

| `@ngx-formly/*` | `6` or `7.1` | Both are supported, see below |

> Using Formly? `@siemens/element-ng/formly` is now based on ngx-formly v7.1. To stay on v6, keep `@ngx-formly/*` at 6 and import `SiFormlyModule` from `@siemens/element-ng/formly-legacy`. To move to v7.1, use the standalone `SiFormlyComponent` from `@siemens/element-ng/formly`. See [dynamic forms](../architecture/dynamic-forms.md).

It sits there because it is a version decision you make while aligning peer dependencies. Both paths are supported, so this must not read as a mandatory v7.1 bump.

docs/update-guides/update-v51.md
kfenner: Reworded so it says where you are and why the section exists:

> Your application builds again. What remains is visual â€” the compiler will not point you to it:

Each item below it now also says what changed, rather than only telling you to go look. Typography, for example:

> **Typography.** Headings are now semibold and font sizes follow a new 1.2 step scale. Use the [page header](../fundamentals/layouts/header.md); the title must be `h2.si-layout-title`.

docs/update-guides/update-v51.md
kfenner: You're right, that was wrong. Rewrote the intro:

> Element v51 runs on Angular 22 and brings in the system tokens of the Siemens Design Language. Your components mostly keep their look. Typography is the change you will notice. The legacy utilities and Sass variables are deprecated, not removed, so they keep working. Only `text-primary` and `btn-ghost` changed their meaning.

Checked against the changelog: legacy `bg-*`/`text-*`, the Bootstrap shadow/elevation APIs and list-group are deprecated but still work. The only two utilities that actually changed meaning are `text-primary` (now accent) and the `btn-ghost` pair. Typography is the one genuinely visible change: headings semibold, new 1.2 step scale.

Dropped the "API removals still fail the build" sentence entirely. That information now sits in "Changes you apply yourself", which says explicitly which rows the compiler catches.

docs/update-guides/update-v51.md
kfenner: Conclusive now, and the intro says so: "It applies all of the following changes, so you do not have to:". I went through all 10 rules in `migrate-to-v51.ts` plus their migration data and listed every one, including the missing-translation handler that the first version had missed.

Also made every bullet say what actually happens. They are grouped by verb so the list stays scannable: Renames / Moves / Replaces / Adds / Removes. For example, instead of "Migrates `si-split`":

> Replaces `scale`, `[sizes]`, `collapseDirection` and `showCollapseButton` on `si-split` with `size`, `unit` and `collapsible`

and the added attributes now say which component they land on and why:

> Adds `listWidthUnit="fr"` to resizable `si-list-details`, and `mainContainerWidthUnit="fr"` to resizable `si-main-detail-container`, so your layout stays as it is

One thing this pass uncovered: the schematic deletes `uploadTextFileSelect` but cannot move its text, so the file uploader button label silently gets shorter. That is now called out on the bullet and has its own row in the manual table.

docs/update-guides/update-v51.md
```diff
@@ -0,0 +1,134 @@
+# Update to Element v51
+
+Element v51 runs on Angular 22 and brings in the system tokens of the Siemens Design Language.
```
kfenner: Requires is the better term here

```suggestion
Element v51 requires Angular 22 and brings in the system tokens of the Siemens Design Language.
```

kfenner: Applied. Checked `projects/element-ng/package.json` to be sure the stronger wording holds: the peer dependency is `"@angular/core": "22"`, an exact major, so `requires` is the accurate term.

docs/update-guides/update-v51.md
```diff
@@ -5,8 +5,7 @@ Most Element components keep their appearance. Typography is the main visual cha
 utilities and Sass variables are deprecated, but they still work. Only `text-primary` and
 `btn-ghost` changed meaning.
 
-The `migration-v51` schematic applies most of the breaking changes. This guide covers the
-remaining steps.
+The `migration-v51` schematic applies most breaking changes; this guide covers what it does not.
```
kfenner: nah, prev version was better.

docs/update-guides/update-v49.md
spike-rabbit, line 1: I would be in favor of keeping this.
Otherwise it is hard to find and I do not see a big advantage of dropping it.

docs/update-guides/update-v51.md
```diff
@@ -0,0 +1,138 @@
+# Update to Element v51
+
+Element v51 requires Angular 22. It introduces the system tokens of the Siemens Design Language.
```
spike-rabbit, line 3: I think we should add a note here, that 50 does not exists

spike-rabbit (changes requested):

docs/update-guides/update-v49.md
kfenner: last time you removed v48 if I'm not mistaken, but sure, we can keep them👍

kfenner: Kept. The v49 guide stays in this version of the docs, listed under Update guides next to v51. The v51 guide now links to that page.

docs/update-guides/update-v51.md
kfenner: Added. The intro now states that there is no Element v50.

kfenner:
@spike-rabbit @dr-itz @chintankavathia Update guide is ready, feedback is welcome 🙏

docs/update-guides/update-v51.md
```diff
@@ -0,0 +1,184 @@
+# Update to Element v51
+
+!!! info "Ehm… what happened with Element v50?"
```
spike-rabbit, line 3: For the blog posts this is cool, but here it feels slightly off, as it does provides a nice soft introduction. There is very hard cut in language afterwards.

docs/update-guides/update-v51.md
```diff
@@ -0,0 +1,184 @@
...
+remaining steps.
+
+## Before you start
+
+1. Update to the latest Element v49 release.
+   If you are updating from v48 or older, follow the [v49 update guide](update-v49.md) first.
+2. Commit or stash your changes. The update rewrites templates, styles, and TypeScript across your project.
+3. Remove all deprecated Element APIs. Use [@typescript-eslint/no-deprecated](https://typescript-eslint.io/rules/no-deprecated/) to find them.
+
+## Update Angular and peer dependencies
```
spike-rabbit: In the past we encouraged applications to split the Angular update from Element. So this shoud be part of Before you start. Dependency updates that must be connected to Element (cannot not be done together with the Angular update, so angular/aria, map-styles and girdstack bump should come later.
ngx-translate core should before only be bumped to 17, datatable to 26, the newer versions are only supported by Element 51

docs/update-guides/update-v51.md
```diff
@@ -0,0 +1,184 @@
...
+Rename the remaining class names yourself, using the same mappings the schematic applied to your
+templates. Most mappings are one-to-one. Two pairs swapped meaning, so apply those in this order:
+
+1. `text-primary` → `text-accent`, then `text-body` → `text-primary`
+2. `btn-ghost` → `btn-tertiary-ghost`, then `btn-primary-ghost` → `btn-ghost`
+
+Limit the search to the places listed above, because a project-wide find and replace would also
+rename templates the schematic already migrated.
+
+## Fix remaining API changes
```
spike-rabbit:
```suggestion
### Fix remaining API changes
```

spike-rabbit (changes requested):
After reading it again, I think there are a few things off / wrong.

docs/update-guides/update-v51.md
kfenner: I really like it, can we keep it anyway? 😸

kfenner: I mean, it's a bit of fun, I'd hope nobody is really distracted/confused by this.

docs/update-guides/update-v51.md
```diff
@@ -0,0 +1,184 @@
...
+
+The `migration-v51` schematic applies most of the breaking changes. This guide covers the
+remaining steps.
+
+## Before you start
+
+1. Update to the latest Element v49 release.
+   If you are updating from v48 or older, follow the [v49 update guide](update-v49.md) first.
+2. Commit or stash your changes. The update rewrites templates, styles, and TypeScript across your project.
+3. Remove all deprecated Element APIs. Use [@typescript-eslint/no-deprecated](https://typescript-eslint.io/rules/no-deprecated/) to find them.
```
chintankavathia: This is not necessarily applicable to everything because we have some deprecated stuff which are there for a very long time e.g MenuItemLegacy

docs/update-guides/update-v51.md
```diff
@@ -0,0 +1,184 @@
...
+```
+
+Review the result in the running application. Consult UX designers where a mapping depends on
+context. Visual regression tests help catch mistakes in this migration.
+
+## Review the UI
+
+The changes below are visual, so check them in the running application:
+
+**Typography.** Headings are now semibold, and font sizes follow a new 1.2 step scale. For page
```
chintankavathia: Not sure but should there be fullstops with every bullets here?

chintankavathia (dismissed):

docs/update-guides/update-v51.md
kfenner: Restructured as suggested.

Angular v22 is now step 2 of *Before you start*, with an explicit note to keep it separate from the Element update.

I dropped the note about stopping at `@ngx-translate/core` v17 and `@siemens/ngx-datatable` v26 during that step. The guide never asks you to touch those packages there, and Element v49 pins them to `16 - 17` and `22 - 26`, so bumping too far fails the install with a peer conflict rather than breaking silently. The peer dependency table is right where you need it once you get to the Element update.

That table moved under *Update Element* and now leads with the three that are genuinely tied to v51 (`@angular/aria`, `@siemens/map-styles`, `gridstack` v13), followed by the three where the version you already run keeps working. A sentence after the table spells that out so nobody bumps translate to v18 or datatable to v27 without wanting to.

I also made it explicit that `ng update` bumps the `@siemens/*` packages via its package group but does not install peer dependencies.

docs/update-guides/update-v51.md
kfenner: Accepted, the heading is now `###` so it stays inside the Element update section.

docs/update-guides/update-v51.md
kfenner: How would you write it?

docs/update-guides/update-v51.md
kfenner: true, this is weird.

docs/update-guides/update-v51.md
kfenner: Good catch, all was wrong. Reworded to drop it and explain why the step is still worth doing:

> 3. Clean up deprecated Element APIs.
>    Use [@typescript-eslint/no-deprecated](https://typescript-eslint.io/rules/no-deprecated/) to find them.
>    Not every deprecated API is removed in v51, but the ones that are will turn into compile errors
>    after the update. Handling them now keeps that list short.

docs/update-guides/update-v51.md
kfenner: Switched them to a colon (`**Typography:**`), which is what the rest of the docs use: 120 occurrences of `**Label:**` against 6 of `**Label.**`, and the 6 were all on this page.

Kept them as paragraphs rather than a list. Three of the five run to two or three sentences and the last one carries a code block, so a list would nest awkwardly without making the scanning any easier. Happy to switch if you prefer the list look.

spike-rabbit (approved):
👍

PR #2726, opened 2026-09-04
docs: make version selector keyboard accessible

kfenner:
Follow-up to #2723. The click-only override also blocked the keyboard: Material opened the menu on `:hover` / `:focus-within`, and disabling that left no way to Tab into the list.

The version control is a small disclosure (not a combobox):

- Native button with `aria-expanded`, `aria-controls`, and `aria-label` (`Select version, current: …`)
- Real links in a `ul`, `hidden` when closed, `aria-current="page"` on the current version
- Click to toggle; hover does not open
- Tab through the links when open; ArrowUp/Down open and focus the current version; Escape closes and returns focus
- Close on `document` `focusin` (after focus has left) so Chrome does not hide the list while Tabbing, and Safari click is not cancelled by a null `relatedTarget`
- List sits flush under the button (`top: 100%`) so a second click still hits the toggle

Tested locally with `curl …/versions.json -o docs/versions.json` then `pnpm docs:serve`, in Chrome and Playwright WebKit. Real Safari still needs a Mac check.

Backports to v49 and v48 are intentionally not included yet.

---

[Documentation](https://d33c9dcnqinn2a.cloudfront.net/pr-2726/pages/).
[Examples](https://d33c9dcnqinn2a.cloudfront.net/pr-2726/pages/element-examples/).
[Dashboards Demo](https://d33c9dcnqinn2a.cloudfront.net/pr-2726/pages/dashboards-demo/).
[Playwright report](https://d33c9dcnqinn2a.cloudfront.net/pr-2726/playwright-report/).

**Coverage Reports:**

Code Coverage

- [charts-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-2726/pages/coverage/charts-ng/)
- [dashboards-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-2726/pages/coverage/dashboards-ng/)
- [element-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-2726/pages/coverage/element-ng/)
- [element-translate-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-2726/pages/coverage/element-translate-ng/)
- [maps-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-2726/pages/coverage/maps-ng/)
- [native-charts-ng](https://d33c9dcnqinn2a.cloudfront.net/pr-2726/pages/coverage/native-charts-ng/)

PR #2804, opened 2026-09-21
docs: make version selector keyboard accessible

kfenner:
Backport of the version selector from main (click + keyboard + a11y), see #2726.

Open the docs version menu on click or keyboard, not hover.
Tab through the version links; ArrowUp/Down focus the current
version; Escape closes. Close on document focusin so Chrome
does not hide the list while tabbing and Safari does not
cancel the click. Name the control with an accessible label.
Keep Material's header height so the label stays vertically
centered, and inset the list-item focus ring.

Related to #2726

---


chintankavathia (approved):

chintankavathia (changes requested):
@kfenner any idea why it updated package-lock here?

kfenner:
> @kfenner any idea why it updated package-lock here?

@chintankavathia Build fails without the updates in the package.lock. See the commit message:

```
Add missing optional encoding/iconv-lite entries so `npm ci --include=optional` works on current npm 11.
```

chintankavathia (approved):

spike-rabbit (approved):

